PR #1, opened 2016-07-01
Simplify and fix IO

kevinmehall:
Tested example with tcp, udpout, and udpin.


tcr3dr:
I'm actually not going to be online for the next eight days (yikes) so I just gave this a cursory review and it looks good. Might want to get @mrpollo to be able to reproducibly test it and then approve its merge.

src/connection.rs
```diff
@@ -91,370 +14,160 @@ pub fn parse_mavlink_string(buf: &[u8]) -> String {
...
-                // state is used to determine whether we are currently reading
-                // or writing.
-                // self.reregister(event_loop);
             }
-            _ => panic!("Received unknown token"),
+            
+            if let Ok((_, m)) = read(&mut lock.recv_buf) {
+                return Ok(m);
+            } else {
+                // Drop the remainder of the packet
```
tcr3dr: Note: we probably want to comment that this won't work if we're processing UDP packets packed with several MAVLink messages as solo has outputted (externally) in the past. I can't see any scenario where it impacts our driver code today though.

kevinmehall: No, that case should work (but needs testing), because it will return Ok, and on the next call, it will re-use the buffer instead of receiving a new packet. This branch is to clear out garbage at the end of the packet.

Assuming the only errors we can get from IO on the Cursor are UnexpectedEOF, this is not strictly necessary, though, because the check for 0xfe ensures forward progress.

PR #4, opened 2016-08-26
Fix CRCs on written packets

kevinmehall:


liamstask:
lgtm. would be nice to add a roundtrip unit test, but merging for now.

PR #7, opened 2016-11-18
Add bindings for ardupilotmega.xml 

frizensami:
This patch adds bindings for messages defined in ardupilotmega.xml. XML data was manually copied and pasted into combined.xml - perhaps this could be automated in the future.

### DIFF
--- a/ardupilotmega.xml
+++ b/ardupilotmega.xml
@@ -1,0 +1,1564 @@
+<?xml version='1.0'?>
+<mavlink>
+    <include>common.xml</include>
+    <!-- note that APM specific messages should use the command id range from 150 to 250, to leave plenty of room for growth of common.xml If you prototype a message here, then you should consider if it is general enough to move into common.xml later -->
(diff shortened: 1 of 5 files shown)


frizensami:
wat

Been testing it with Arducopter software-in-the-loop: seems to be detecting the new messages appropriately.

podhrmic:
Probably obsolete, as it is trivial to add additional xml files to parse, but it is unclear whether it is something we should do.

hamishwillee:
Hi @podhrmic The general model for all our other generators is that they take mavlink version and one or more dialects as inputs, and generate libraries for the specified dialect.

For pre-built versions the model (at least for C and Python) is to generate libraries for all dialects in mavlink/mavlink and include them in the package.

While I don't understand what this code is doing, I suggest you try emulate that approach.

laptou:
This PR is covered by #40, consider closing it when 40 is merged? @patrickelectric

PR #8, opened 2017-05-20
Fix array reader codegen

kersny:
Fixes the array reading code generator to use writeln! rather than println!, so that messages with arrays are properly parsed.


podhrmic:
Should be fixed in #13

PR #9, opened 2018-12-11
Add support for serial connection

podhrmic:
Now it is possible to read Mavlink packets directly off a serial port.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 target
 Cargo.lock
+.cproject
+.project
+.settings/*
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,2 @@
 target
 Cargo.lock
-.cproject
-.project
-.settings/*
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -1,11 +1,13 @@
 use common::MavMessage;
-use {Header, read, write};
+use {read, write, Header};
 
 use std::sync::Mutex;
-use std::net::{TcpStream, UdpSocket, ToSocketAddrs, SocketAddr};
+use std::net::{SocketAddr, TcpStream, ToSocketAddrs, UdpSocket};
 use std::io::{self, Read};
 
 use std::str::FromStr;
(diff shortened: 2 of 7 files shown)


podhrmic:
@bekicot  Yes, this is supposed (and was tested on Linux) to work with Pixhawk connected over USB-to-UART converter. I would be curious if you can test it on Mac.

bekicot:
@podhrmic Sure, I'll be back in several days, the device along with the drone is at the office.

.gitignore
```diff
@@ -1,2 +1,5 @@
 target
 Cargo.lock
+.cproject
+.project
+.settings/*
```
kevinmehall: If these are for your IDE, they should go in your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) rather than adding them to every repo you use.

src/connection.rs
```diff
@@ -200,17 +214,76 @@ impl MavConnection for Tcp {
...
+    }
+}
+
+impl MavConnection for Serial {
+    fn recv(&self) -> io::Result<MavMessage> {
+        let mut port = self.port.lock().unwrap();
+
+        loop {
+            if let Ok((_, m)) = read(&mut *port) {
+                return Ok(m);
```
kevinmehall: Won't this infinite loop on any error that happens continuously, for instance if you unplug the USB-serial adapter? It should propagate errors to the caller.

.gitignore
podhrmic: Good point, will fix it.

src/connection.rs
podhrmic: I haven't tested it extensively, but yes, the errors should be propagated. Will fix it.

podhrmic:
@kevinmehall I believe your suggestions were addressed with the latest commit. let me know if it there is anything else you want me to address before the merge.

podhrmic:
Closing in favor of #13

Issue #12, opened 2019-01-25
TCP receive blocks on OSX: MavConnection::recv never returns messages


tstellanova:
Testing this library with a TCP connection, send seems to work but receive does not. 
The problem seems to be that the socket read has no timeout by default, so the read waits forever. 

Setting the timeout:
```
sock_clone.set_read_timeout(Some(Duration::from_millis(100))).unwrap();
```
and then loop-ignoring the resulting `io::ErrorKind::WouldBlock` seems to unstick the socket.  I've got a WIP fork for this -- if this repo is still maintained, I can submit a PR for that when it's ready.

podhrmic:
@tstellanova I believe this should be fixed in #13 ?

tstellanova:
Yes, fixed in #13

tstellanova:
Fixed

PR #13, opened 2019-02-08
Major refactor

podhrmic:
This is a major overhaul of rust-mavlink, probably leading towards a new version on crates.io. 

@podhrmic :
- Changed code generation to use `quote!` procedural macro
- Generate proper enums in message fields (using `bitflags!`macro)
- Added Serial connection
- Separate `std` and `no-std` build (to enable embedded targets in the future)
- Update mavlink.xml

@tstellanova :
- added TCP connections
- added mavlink v2 support and tests
- added support for truncated payloads


Cargo.toml
```diff
@@ -1,20 +1,43 @@
...
 name = "mavlink"
 version = "0.4.2"
-authors = ["Tim Ryan"]
+authors = ["Todd Stellanova", "Michal Podhradsky", "Kevin Mehall", "Tim Ryan"]
 build = "build/main.rs"
-description = "Parses the MAVLink data interchange format for UAVs."
+description = "Implements the MAVLink data interchange format for UAVs."
 license = "MIT/Apache-2.0"
-repository = "https://github.com/3drobotics/rust-mavlink"
+repository = "https://github.com/tstellanova/rust-mavlink.git"
```
kevinmehall: Let's not change this until we've found a permanent home for the repo.

build/main.rs
```diff
@@ -17,5 +20,8 @@ pub fn main() {
     let dest_path = Path::new(&out_dir).join("common.rs");
     let mut outf = File::create(&dest_path).unwrap();
 
-    parser::generate_mod(&mut inf, &mut outf);
+    parser::generate(&mut inf, &mut outf);
+
+    // Dont run build.rs unless it is changed
+    println!("cargo:rerun-if-changed=build.rs");
```
kevinmehall: Cargo always re-runs if `build.rs` has changed. However, you should put `common.xml` here.

build/parser.rs
```diff
@@ -1,61 +1,542 @@
...
+            .collect::<Vec<Tokens>>();
+        (field_toks, encoded_payload_len)
+    }
+
+    /// Generate description for the given message
+    #[cfg(feature = "emit-description")]
+    fn emit_description(&self) -> Tokens {
+        let mut desc = String::from(format!("\n/// id: {}\n", self.id));
+        if let Some(val) = self.description.clone() {
+            desc = desc + &format!("/// {}.\n",val);
```
kevinmehall, line 344: It would be better to expand to `#[doc = "..."]` for compatibility with token trees and to be less sensitive to whitespace, and you can generate it with `quote!()`.

build/parser.rs
```diff
@@ -77,37 +558,112 @@ pub enum MavType {
...
+            Int8 => quote!{#val = #buf.get_i8();},
+            Int16 => quote!{#val = #buf.get_i16_le();},
+            Int32 => quote!{#val = #buf.get_i32_le();},
+            Int64 => quote!{#val = #buf.get_i64_le();},
+            Float => quote!{#val = #buf.get_f32_le();},
+            Double => quote!{#val = #buf.get_f64_le();},
+            Array(t, size) => {
+                if size > 32 {
+                    // it is a vector
+                    let r = t.rust_reader(Ident::from("let val"), buf.clone());
```
kevinmehall, line 607: Should pass `quote!(let val)` and change the argument type instead of wrapping strings in an `Ident`.

Better yet, just return the expression tokens and leave the assignment to the caller?

build/parser.rs
```diff
@@ -147,12 +705,14 @@ impl MavType {
         }
     }
 
+    /// Return rust equivalent of a given Mavtype
+    /// Used for generating struct fields.
     pub fn rust_type(&self) -> String {
```
kevinmehall, line 702: Should return `Tokens` instead of `String`.

build/parser.rs
```diff
@@ -305,32 +865,56 @@ pub fn parse_profile(file: &mut Read) -> MavProfile {
...
+                                            v.into_iter().collect()
+                                        })
+                                        .collect::<Vec<String>>()
+                                        .join("");
+                                        */
                                     message.name = attr.value.clone();
                                 }
                                 "id" => {
-                                    message.id = attr.value.parse::<u8>().unwrap();
+                                    //message.id = attr.value.parse::<u8>().unwrap();
```
kevinmehall, line 908: Is this supposed to be behind a `cfg(not(feature = "mavlink2"))` or just obsolete and can be removed?

src/connection.rs
```diff
@@ -1,216 +1,439 @@
...
+
+        loop {
+            match read(&mut *port) {
+                Ok((_, m)) => {
+                    return Ok(m);
+                }
+                Err(e) => {
+                    println!("{:?}",e);
+                    match e.kind() {
+                        io::ErrorKind::UnexpectedEof => {
```
kevinmehall: I'd rather see a whitelist of errors that it retries on, and propagate anything else, to make it less likely to infinite-loop if something weird happens. These kinds of errors are often inconsistent between OSs.

src/connection.rs
```diff
@@ -1,216 +1,439 @@
...
 use std::str::FromStr;
 
+#[cfg(feature = "tcp")]
+use std::net::{TcpListener, TcpStream};
+#[cfg(feature = "tcp")]
+use std::time::Duration;
+
+#[cfg(feature = "serial")]
+extern crate serial;
+#[cfg(feature = "serial")]
```
kevinmehall: Consider splitting udp, tcp, and serial into separate files, and putting the `#[cfg]` on e.g. the `mod udp;` import instead of every single item in this file.

src/connection.rs
```diff
@@ -1,216 +1,439 @@
...
+                Ok( (header, msg) ) => {
+                    return Ok((header, msg) );
+                },
+                Err(e) => {
+                    match e.kind() {
+                        io::ErrorKind::WouldBlock => {
+                            //println!("would have blocked");
+                            continue;
+                        },
+                        _ => panic!("TCP read_msg error: {}", e),
```
kevinmehall: You should propagate the error instead of panicking.

src/connection.rs
```diff
@@ -1,216 +1,439 @@
...
+        loop {
+            let mut lock = self.reader.lock().expect("tcp read failure");
+            match read_msg(&mut *lock) {
+                Ok( (header, msg) ) => {
+                    return Ok((header, msg) );
+                },
+                Err(e) => {
+                    match e.kind() {
+                        io::ErrorKind::WouldBlock => {
+                            //println!("would have blocked");
```
kevinmehall: What's the purpose of retrying here? It shouldn't get a `WouldBlock` unless someone set the socket to non-blocking mode, in which case they probably want the `WouldBlock` error to integrate with some kind of async system, instead of continuously retrying in a tight loop until a packet comes in. Did you mean `io::ErrorKind::Interrupted`?

src/lib.rs
```diff
@@ -19,55 +50,282 @@ pub mod common {
...
+pub const MAV_STX: u8 = 0xFE;
+
+/// Message framing marker for mavlink v2
+pub const MAV_STX_V2: u8 = 0xFD;
+
+
+impl MavHeader {
+    /// Return a default GCS header, seq is replaced by the connector
+    /// so it can be ignored. Set `component_id` to your desired component ID.
+    pub fn get_default_header() -> MavHeader {
```
kevinmehall, line 69: Consider just implementing the `Default` trait with this implementation. It could be confusing that this differs from the `Default::default()` you currently `#[derive]` which zeros all the fields.

src/lib.rs
```diff
@@ -19,55 +50,282 @@ pub mod common {
...
+
+        // serialize header
+        v.push(self.header.system_id);
+        v.push(self.header.component_id);
+        v.push(self.header.sequence);
+
+        // message id
+        #[cfg(feature="mavlink2")]
+        {
+            let bytes: [u8; 4] = unsafe { transmute(self.msg.message_id().to_le()) };
```
kevinmehall:
```suggestion
            let bytes: [u8; 4] = self.msg.message_id().to_le_bytes();
```
(newly stabilized in Rust 1.32)

src/lib.rs
```diff
@@ -19,55 +50,282 @@ pub mod common {
...
+        msgid_buf[0] = r.read_u8()?;
+        msgid_buf[1] = r.read_u8()?;
+        msgid_buf[2] = r.read_u8()?;
+
+        let header_buf = &[payload_len as u8,
+            incompat_flags, compat_flags,
+            seq, sysid, compid,
+            msgid_buf[0],msgid_buf[1],msgid_buf[2]];
+
+        let msgid: u32 = unsafe { transmute(msgid_buf) };
```
kevinmehall:
```suggestion
        let msgid: u32 = u32::from_le_bytes(msgid_buf);
```

src/lib.rs
```diff
@@ -19,55 +50,282 @@ pub mod common {
...
+    /// Return the frame header
+    pub fn header(&self) -> MavHeader {
+        self.header
+    }
 }
 
-/// Read a MAVLink message from a Read stream.
-pub fn read<R: Read>(r: &mut R) -> io::Result<(Header, MavMessage)> {
+/// Read a MAVLink v1  message from a Read stream.
+#[cfg(all(feature = "std", not(feature="mavlink2")))]
```
kevinmehall: The `mavlink2` cfg flag shouldn't remove runtime support for Mavlink 1. Cargo feature flags have to be purely additive because if multiple dependencies of a program transitively depend on this one, it gets compiled with the union of their cargo features.

Cargo.toml
```diff
@@ -1,20 +1,43 @@
...
 
 [build-dependencies]
 crc16 = "0.3.3"
-byteorder = "0.5.3"
+bytes = "0.4"
 xml-rs = "0.2"
+quote = "0.3"
```
kevinmehall, line 15: Be advised that quote 0.3 was last updated two years ago and differs significantly from 0.6.

You could make an argument that 0.3 is better for this use case because it doesn't need to integrate with proc-macro and lets you get away with abusing `Ident` into a way of pasting in arbitrary strings, but know that it's outdated and unmaintained.

build/parser.rs
```diff
@@ -1,61 +1,542 @@
...
+        println!("Updating messages");
+        let msgs = self.messages.into_iter().filter(|x| x.id <= 254).collect::<Vec<MavMessage>>();
+        self.messages = msgs;
+        self
+    }
+
+    /// Simple header comment
+    fn emit_comments(&self) -> Ident {
+        Ident::from(format!(
+            "// This file was automatically generated, do not edit \n"
```
kevinmehall, line 57: You don't even need to misuse `Ident` for this, just write it to the output file before the source code.

build/parser.rs
```diff
@@ -339,12 +923,31 @@ pub fn parse_profile(file: &mut Read) -> MavProfile {
...
+                                            .split("_")
+                                            .map(|x| x.to_lowercase())
+                                            .map(|x| {
+                                                let mut v: Vec<char> = x.chars().collect();
+                                                v[0] = v[0].to_uppercase().nth(0).unwrap();
+                                                v.into_iter().collect()
+                                            })
+                                            .collect::<Vec<String>>()
+                                            .join(""),
+                                    );
```
kevinmehall, line 938: This expression is duplicated twice above (one commented). Make it a function instead?

build/parser.rs
```diff
@@ -305,32 +865,56 @@ pub fn parse_profile(file: &mut Read) -> MavProfile {
...
+                                        .split("_")
+                                        .map(|x| x.to_lowercase())
+                                        .map(|x| {
+                                            let mut v: Vec<char> = x.chars().collect();
+                                            v[0] = v[0].to_uppercase().nth(0).unwrap();
+                                            v.into_iter().collect()
+                                        })
+                                        .collect::<Vec<String>>()
+                                        .join("");
+                                        */
```
kevinmehall, line 904: Remove commented code

src/lib.rs
```diff
@@ -19,55 +50,282 @@ pub mod common {
...
+                MavHeader {
+                    sequence: seq,
+                    system_id: sysid,
+                    component_id: compid,
+                },
+                msg,
+            ));
+        }
+        else {
+            println!("invalid MavMessage::parse");
```
kevinmehall: Maybe return ErrorKind::InvalidData. Library code in general shouldn't just print its own error messages. Let the program using this library control its UI.

src/lib.rs
```diff
@@ -19,55 +50,282 @@ pub mod common {
...
+        let crc = r.read_u16::<LittleEndian>()?;
 
         let mut crc_calc = crc16::State::<crc16::MCRF4XX>::new();
         crc_calc.update(&[len as u8, seq, sysid, compid, msgid]);
         crc_calc.update(payload);
         crc_calc.update(&[MavMessage::extra_crc(msgid)]);
-        if crc_calc.get() != crc {
+        let recvd_crc = crc_calc.get();
+        if recvd_crc != crc {
+            println!("msg id {} len {} , crc got {} expected {}", msgid, len, crc, recvd_crc );
```
kevinmehall: `println!()` in library code is discouraged.

build/parser.rs
```diff
@@ -1,61 +1,542 @@
...
+                    #(#deser_vars)*
+                    Some(_struct)
+                }
+            }
+    }
+
+    fn emit_rust(&self) -> Tokens {
+        let msg_name = self.emit_struct_name();
+        let (name_types, msg_encoded_len) = self.emit_name_types();
+        let payload_len_desc = Ident::from(format!("pub const ENCODED_LEN: usize = {};", msg_encoded_len) );
```
kevinmehall: Use `quote!()`.

Cargo.toml
tstellanova: good catch

src/connection.rs
tstellanova: will do

src/connection.rs
tstellanova: This was a workaround for #12 -- I'll try another option.

src/lib.rs
tstellanova: The comment about additive features is fair.  @podhrmic we should talk about a strategy for supporting both mavlink1 and mavlink2.  Other implementations support either one or the other... I think at least one implementation supports negotiation (or fallback to v1).

Cargo.toml
tstellanova: @podhrmic I forget whether I introduced this in an effort to move to stable.  Do you want to tackle this?

tstellanova:
Thanks for the feedback @kevinmehall  -- @podhrmic I'll try to submit some updates for the noted changes by Monday.  Let me know if you have ideas on how to support fallback from v2->v1 (with the option to remove feature flags for either).

hamishwillee:
@kevinmehall Yes, 
- a compliant MAVLink 2 implementation must support MAVLink 1 too
- magic/protocol maker can be used to differentiate between types (as per C [docs here](https://mavlink.io/en/mavgen_c/))

Just FYI, the expect version handshaking and version negotiation is [documented here](https://mavlink.io/en/guide/mavlink_version.html#version_handshaking)

tstellanova:
> @kevinmehall Yes,
> 
> * a compliant MAVLink 2 implementation must support MAVLink 1 too
> * magic/protocol maker can be used to differentiate between types (as per C [docs here](https://mavlink.io/en/mavgen_c/))
> 
> Just FYI, the expect version handshaking and version negotiation is [documented here](https://mavlink.io/en/guide/mavlink_version.html#version_handshaking)

Thanks for the additional info @hamishwillee .  FYI we have a fix in progress that will support either mavlink1 and mavlink2 channels in this rust library, selectable at runtime.   If I understand the intent of the handshaking docs, it appears handshaking is implemented at the application level (ie a GCS application _may_ send `MAV_CMD_REQUEST_PROTOCOL_VERSION` and a Vehicle application _may_ respond). Since this library is agnostic as to where it is used, it will be up to the apps that use this library to implement handshaking if they choose, switching the channel dynamically to v1->v2.

hamishwillee:
> Since this library is agnostic as to where it is used, it will be up to the apps that use this library to implement handshaking if they choose, switching the channel dynamically to v1->v2.

Yes, that is all correct! Thanks.

podhrmic:
Added @tstellanova 's changes

PR #18, opened 2019-10-12
Move to Rust 2018 edition and use serde serialization

patrickelectric:
- Move to Rust 2018 edition, helping the integration with new crates and some [general corrections](https://github.com/rust-lang-nursery/edition-guide/blob/5521dada71a232403bcce6870eed92897590561e/src/rust-2018/edition-changes.md).
- Use serde serialization to allow generic and more common data access in a generic format.
    - The problem with the messages defined inside enums does not allows the users to use the library in a generic way, resulting in a match/case for each enum for each message to access content. This makes any kind of generic programming almost impossible without hacks around the types and generated code.
    - To fix this problem, this patchs uses [serde crate](https://crates.io/crates/serde), that helps with the transformation between the most popular formats:

image

ref: [serde.rs](https://serde.rs)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ description = "Implements the MAVLink data interchange format for UAVs."
 readme = "README.md"
 license = "MIT/Apache-2.0"
 repository = "https://github.com/mavlink/rust-mavlink"
+edition = "2018"
 
 [build-dependencies]
 crc16 = "0.3.3"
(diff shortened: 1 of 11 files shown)


patrickelectric:
ping @kevinmehall @tstellanova

kevinmehall:
Some late feedback: You can and should make serde an optional dependency, as it's not relevant to most users.

PR #21, opened 2019-12-28
Parse: Flags: Improve expected message and force bits validation 

patrickelectric:
Some vehicles can set and send bits that are not used/declared in the message definition.


podhrmic:
Looks good to me.

PR #30, opened 2020-01-23
parser: Deal with extensions keys

patrickelectric:
Add feature emit-extensions to allow builds with such messages, by default we ignore it 

Improve support for other dialects

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ optional = true
 
 [features]
 "emit-description" = []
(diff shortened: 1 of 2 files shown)


patrickelectric:
@michael-p tested and it fixes his problem.

patrickelectric:
After this PR, I'll send a second one based over this to fix the wip messages, and after that, I'll integrate all protocol definition files to help with the first stable release.

Issue #14, opened 2019-05-28
Getting ready for an official release


podhrmic:
This is a tracking issue to make rust-mavlink an officially supported Mavlink implementation.

What needs to happen?

- [x] Split messages into one file per message (similar to what mavgen does)
- [x] Add quickcheck tests for each message
- [x] change enums to uints
- [x] Add fuzz testing
- [ ] (optional) Build actual enums on top of message enums
- [ ] Mavlink 2 support, including signing
- [x] Quote update
- [x] Somehow ensure max vector length for messages like v2_extension and statustext
- [ ] Support for other dialects

podhrmic:
FYI work in progress, check out https://github.com/GaloisInc/rust-mavlink/tree/fuzz_testing for the latest commits

podhrmic:
Added fuzz testing, and started adding quickcheck tests.

Next steps:
- figure out how to implement `Arbitrary` for `ArrayVec`
- implement `Arbitrary` for `MavMessage` and `MavFrame` to complete quickcheck tests

@hamishwillee do you think we need to have hardcoded test vectors as well? Or is a random (quickcheck) testing considered sufficient? Meaning - I generate a message (let's say `HEARTBEAT`) with random values and try to serialize and then deserialize it?

hamishwillee:
@podhrmic Sorry for the delay, I'm travelling.

> do you think we need to have hardcoded test vectors as well? Or is a random (quickcheck) testing considered sufficient? Meaning - I generate a message (let's say HEARTBEAT) with random values and try to serialize and then deserialize it?

Short version ... no.

Generally the steer from @peterbarker on the mavgen library has been "try to test to the same level as the other libraries" - ie C. I believe the C library tests every message but I don't know what values are passed in - ie whether they are random or not. 
My gut feeling is that random is good, but we may need to tune this to avoid false positives for some messages. 

What we _should_ do is work out a minimum test strategy for any generator library and document it - so you don't have to ask these questions. Also once we know what is tested, do a gap analysis to see if the coverage is reasonable. @peterbarker Do you have a summary anywhere of what tests are run for pymavlink and the C libraries?

podhrmic:
Thanks @hamishwillee for the update. 

FYI updates on my side if you are interested:
- switching to use `ArrayVec` only for arrays > 32 elements, otherwise will use slices
- working on generating enums as defined in the XML (for easier typechecking) that allow unknown values

podhrmic:
[...]

PR #31, opened 2020-01-25
Use mavlink submodule and remove xml file

### LINKED ISSUE #24: Common message set (common.xml) is missing messages
Dear all,

it seems that the message definition file (`common.xml`) that is used to generate the code in this crate is pretty old, and misses quite a few messages. In my case I stumbled across this because of the `LOGGING_DATA`, `LOGGING_DATA_ACKED` and `LOGGING_ACK` messages (both are Mavlink 2 messages).

It would be great if this crate could be updated to the most recent version!

Thanks
Michael

### LINKED ISSUE #29: Create a new release with the last version of master branch
It would be awesome to have a new release with our last corrections and new message definitions :)

patrickelectric:
Fix #24
Fix #27  
Fix #29


patrickelectric:
ping @tstellanova @podhrmic

patrickelectric:
@podhrmic thanks for the input.

> lease update the README.md to reflect the changes (at least we should mention to do `git submodule init + update`).

I did a small update in main.rs to include the `git submodule update --init`, I believe that changes in README are no more necessary.

> A nice to have feature would be to let the user select if they want to use a different dialect of mavlink (besides common.xml). What do you think?

Yes, this will be the next and hopefully the last PR for the stable release and close #14

PR #38, opened 2020-02-19
MavMessage::parse now maps results instead of unwrapping them

### LINKED ISSUE #37: MavMessage::parse panics on invalid message
https://github.com/mavlink/rust-mavlink/blob/8c43fe7fa2128eaa001c6640fafc746e01a7afd3/build/parser.rs#L195-L202

`MavMessage::parse` is set up to `unwrap()` the value that is provided by `serde`.

Why can't it just return `None` instead of unwrapping? This way, invalid data will not lead to a panic.

laptou:
Now it will not panic if the packet happens to contain invalid or incomplete data. Fixes #37.


patrickelectric:
This looks a interesting approach, but should serde(default) fix that no ?

patrickelectric:
Also @laptou, can you describe what is this fixing ? I can't imagine a scenario where such thing happens, besides a wrong protocol implementation or something. This could also not fail when invalid messages are received, and this can be a bit dangerous to be done in the library.

laptou:
I had a loop in my application that looks like this:

```rust
// initialize packet buffer with zeros
    let mut buf = BytesMut::with_capacity(512);
    buf.resize(512, 0);

    let mut fail_count = 0;

    loop {
        let len = match sock.recv(buf.as_mut()).await {
            Ok(len) => len,
            Err(e) => {
                // if we fail to read from the socket 16 times; exit the loop
                if fail_count < 16 {
                    fail_count += 1;
                    warn!(
                        "failed to read from socket ({:?}), retrying ({:?})",
                        e, fail_count
                    );
                    continue;
                } else {
                    error!("failed to read from socket, giving up");
                    break;
                }
            }
        };

        let (_, msg) = match mv::read_v2_msg(&mut &buf[..len]) {
            Ok(packet) => packet,
            Err(_) => continue, // ignore malformed packets
        };

        // we are only interested in GLOBAL_POSITION_INT and ATTITUDE messages
        match msg {
            MavMessage::GLOBAL_POSITION_INT(..) | MavMessage::ATTITUDE(..) => {
                // Result can be ignored b/c it doesn't matter whether
                // any receivers are out there
                let _ = chan.send(msg);
            }
            _ => {}
        }
    }
```

This is intended to just keep reading packets from the UDP socket (`sock` is a Tokio `UdpSocket`) and send them to another thread using a channel. However, my application seemed to be unstable and would randomly panic after a few minutes when I used this method.

This could also just be that I am reading the data from the socket incorrectly, but I couldn't find any examples that do it differently.

laptou:
Update: I can't seem to reproduce this bug...

patrickelectric:
Hi @laptou,

If you want to create a mavproxy alternative, check my mavlink-proxy: https://github.com/patrickelectric/mavlink-proxy 
I believe that it does what you are looking for.

Issue #35, opened 2020-02-01
Release 0.7 requires recompilation on every build


michael-p:
Dear all,

I just noticed that the recently published version 0.7 is rebuild by Cargo on every run, even though nothing has been modified.

To reproduce, create a new project and add mavlink to it:
```
$ cargo new --bin mavlink-test
$ cd mavlink-test
$ cargo add mavlink
$ cargo build
```
Then, just run the project twice without modifying anything in between:
```
$ cargo run && cargo run
   Compiling mavlink v0.7.0
   Compiling mavlink-test v0.1.0 (/Users/michaelp/Desktop/mavlink-test)
    Finished dev [unoptimized + debuginfo] target(s) in 17.18s
     Running `target/debug/mavlink-test`
Hello, world!
   Compiling mavlink v0.7.0
   Compiling mavlink-test v0.1.0 (/Users/michaelp/Desktop/mavlink-test)
    Finished dev [unoptimized + debuginfo] target(s) in 17.44s
     Running `target/debug/mavlink-test`
Hello, world!
```
Notice that both the mavlink crate as well as the application itself were compiled again.

For comparision, this is the output with version 0.6:
```
$ cargo run && cargo run
   Compiling mavlink-test v0.1.0 (/Users/michaelp/Desktop/mavlink-test)
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/mavlink-test`
Hello, world!
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/mavlink-test`
Hello, world
```

I'm using this crate in a fairly large application where compile times are measured in minutes, so having to rebuild even though nothing has changed is pretty annoying as you can probably imagine :)

Thanks,
Michael

podhrmic:
Hmm, I wonder if this is because we now [checkout](https://github.com/mavlink/rust-mavlink/blob/master/build/main.rs#L19) `common.xml` from mavlink submodule, rather than using [a static copy of the file](https://github.com/mavlink/rust-mavlink/blob/c24eed107cbd8de083948a89c19dc9246dfe18ab/build/main.rs).

We should pin to a specific mavlink commit anyway, so `git checkout ...` would be better. I will see if pinning a specific commit makes this issue go away.

podhrmic:
@michael-p this issue is currently fixed in `master`, and will be shortly published as version `0.8` on crates.io

podhrmic:
[...]

PR #40, opened 2020-03-01
Adding support for other message sets

laptou:
This pull request adds (feature-gated) support for all of the XML message sets defined by MAVProxy. Enabling the feature for a given message set will allow you to parse messages from that message set. Here's how I accomplished this:

## `MavMessage`
Each module in the crate corresponds to a message set, and there is now a `MavMessage` enum inside each module in the crate that represents those messages.

## Includes
Message sets can include others. For example, all of them include `common.xml`, and `ardupilotmega.xml` includes `icarous.xml` and `uavionix.xml` too. In order to represent this, the `MavMessage` enum contains extra variants that wrap messages from other message sets. For example, `mavlink::ardupilotmega::MavMessage` has a variant `mavlink::ardupilotmega::MavMessage::icarous(mavlink::icarous::MavMessage)`.

## `Message` trait
This encapsulates all of the functionality common to all of the message types into a trait. This makes it possible to make types like `MavFrame` and `MavConnection` polymorphic, so that the user can use whichever message type they need.

## Polymorphic `MavConnection`
MavConnection is now parameterized by a type which implements `Message`, which means that when the user creates a `MavConnection`, they will have to decide what kind of messages they are expecting by using a certain `MavMessage` enum. For example, if they are expecting MAVProxy to send ArduPilotMega-flavoured messages, then they should create a `MavConnection<mavlink::ardupilotmega::MavMessage>`.

What do you think? This seems like a big change, so I also bumped the version to 0.8.


laptou:
Note: Travis CI builds are now failing because the `ardupilotmega`, `icarous`, and `uavionix` features need to be enabled in order to compile all of the tests.

patrickelectric:
@laptou can you improve travis to run the tests for each flavor or enabling all ?

laptou:
@patrickelectric Fixed it.

patrickelectric:
You need to rebase this PR over master.  there is some lost commits from master here.

laptou:
Done @patrickelectric

laptou:
By the way, this PR also fixes #35. `mavlink` will still recompile one time due to the XML files being modified by the [patch](https://github.com/mavlink/rust-mavlink/blob/fa685e3b6b2101220ff359ff6f272a4542f5c3d8/build/patches/fix-ekf-status-report.patch), but after that the issue will go away.

Cargo.toml
```diff
@@ -1,8 +1,8 @@
 
 [package]
 name = "mavlink"
-version = "0.7.0"
-authors = ["Todd Stellanova", "Michal Podhradsky", "Kevin Mehall", "Tim Ryan", "Patrick José Pereira"]
+version = "0.8.0"
```
podhrmic, line 4: @patrickelectric should this be a major or minor release?

podhrmic:
Actually, I think it would be good to update the README to reflect this new cool functionality.

Cargo.toml
```diff
@@ -30,13 +29,23 @@ num-derive = "0.2"
...
-[dependencies.byteorder]
-version = "0.5.3"
-optional = true
-
+byteorder = { version = "0.5", optional = true }
 
 [features]
+"ardupilotmega" = []
+"asluav" = []
+"autoquad" = []
```
patrickelectric, line 38: It's missing `common`

patrickelectric: `common` should also be an on by default feature.

This also will fix `cargo test`, it's right now failing, it only works with all features enabled.

Cargo.toml
```diff
@@ -30,13 +29,23 @@ num-derive = "0.2"
...
+"minimal" = []
+"paparazzi" = []
+"python_array_test" = []
+"slugs" = []
+"standard" = []
+"test" = []
+"ualberta" = []
+"uavionix" = []
+"icarous" = []
+
```
patrickelectric, line 50: A final note, will `common::` prefix be everything ? What happens if one is selected over the others or if someone want to change the dialect ?

Cargo.toml
patrickelectric: It depends if it'll break or change the API, from what I can see, the common prefix will be any dialect or all dialects that the feature enable. I believe that the correct way should be a prefix for eat dialect and enable it if such feature is selected.

Cargo.toml
laptou: There's no `common` feature. The common message set can't be disabled, since all other message sets depend on it anyway.

laptou: The reason that the tests fail without features enabled is that I add tests that test whether messages are encoded and decoded properly in a specific (not `common`) message set, so the features that correspond to that message set need to be enabled.

Cargo.toml
laptou: I don't really understand your question. The `common::` prefix only applies to messages and data types that are defined in the common message set.

laptou: The reason that I consider this a breaking change is because some user-facing types, such as `MavConnection<T>` are now generic, where they weren't before.

Cargo.toml
patrickelectric: > There's no `common` feature. The common message set can't be disabled, since all other message sets depend on it anyway.

That's not true, `icarous` and `minimal` does not have `common` in `include` inside the `.xml`.

 

> The reason that the tests fail without features enabled is that I add tests that test whether messages are encoded and decoded properly in a specific (not `common`) message set, so the features that correspond to that message set need to be enabled.

This tests should have `#[cfg(feature = "message_set")]` to only enable with the feature selected.

Cargo.toml
patrickelectric: I see it now, thanks for the explanation!

Cargo.toml
laptou: Ah, I didn't notice this was the case. I've added a `common` feature and feature-gated the tests that require the `ardupilotmega` feature.

laptou:
Everything looks ready to me. @podhrmic @patrickelectric

patrickelectric:
It looks ok to me, but I would recommend to squash some commits to avoid the unnecessary noise in the git history.

PR #41, opened 2020-03-05
travis: Add deploy via cargo with tags

patrickelectric:
@podhrmic Can you create a secure travis token and share it with me to add in the file ?
https://docs.travis-ci.com/user/encryption-keys/

I prefer this online website to do it: https://travis-encrypt.github.io/
but you can use the travis official link to generate it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,9 @@
 language: rust
 cache: cargo
+
+deploy:
+  - provider: cargo
+    token:
+      secure: ${SECURE_KEY}
+    on:
+      tags: true


patrickelectric:
Also, can we do a minor release ?

podhrmic:
I am not sure I understand what are the keys for? So cargo can use travis for testing?

podhrmic:
> Also, can we do a minor release ?

I assume you mean after merging #40? I think we sure can.

patrickelectric:
@podhrmic with the variable it'll deploy the package via cargo to crates.io automatically

podhrmic:
OK, it turns out I don't have permissions to do so:-/ Perhaps @kevinmehall or @tstellanova has the rights?

kevinmehall:
@podhrmic You have the same level of access I do at https://crates.io/crates/mavlink/owners.

If you decide to do this, you should consider making a separate account for it. As far as I know crates.io tokens aren't scoped to a particular package, so it would be able to publish to any of your packages.

podhrmic:
OK, this should be fixed with https://github.com/mavlink/rust-mavlink/commit/c975d078e5444114f3f608dc0cb72e2c177d8b75

I created a new account for that, encrypted the token as mentioned and added it to `.travis.yml`

I still had to manually push a new 0.8 tag.

I will close this PR once travis builds and hopefully pushes the new version to crates.

podhrmic:
0.8 released! The travis->cargo seems to work for tagged releases, but in this case it failed on some internal travis error. So I did the release manually.

PR #43, opened 2020-03-11
Fix build

patrickelectric:
It that I did a sad rebase, this fix the `default_message_from_id` patch and also adds a test for it


patrickelectric:
@podhrmic this fix the build, and will make our goal closer to the 0.8 release

PR #45, opened 2020-03-11
tests: Fix format from patch before format check

patrickelectric:


patrickelectric:
@podhrmic hopefully final commit for the release

PR #49, opened 2020-04-15
Return Results instead of panicking

### LINKED ISSUE #37: MavMessage::parse panics on invalid message

laptou:
Resolves #37. All of the functions that involve parsing a `MavMessage` will now return a `Result<Message, MessageReadError>` instead of panicking upon encountering invalid data.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 
 [package]
 name = "mavlink"
-version = "0.8.1"
+version = "0.8.2"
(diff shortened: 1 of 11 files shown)


podhrmic:
@laptou can you pls fix the build error? Otherwise I am not opposed to changing the return type. Also, you might want to update the version to 0.8.2

laptou:
Sorry, it took me a while to figure out that the reason the build failed was because of formatting errors.

PR #81, opened 2021-01-04
Add support for embedded, no_std environments

KoffeinFlummi:
I needed to use this crate on an STM32F4, so I tweaked the crate to compile without `std`, and added support for the `serial::Read<u8>` and `serial::Write<u8>` traits from `embedded-hal`. I tried to make as few changes to the existing code as possible, so the `std` development is not inconvenienced unnecessarily, however, some changes could not be avoided.

Overview of the changes:
- Updated the `bytes` dependency to 1.0. This allows using it in `no_std` environments, but it did require replacing `Bytes::from(x).into_buf()` with `BytesMut::from(x)` and explicitly converting them to an iterator for collecting them into a `Vec`.
- Enabled the `byteorder` crate for all features, and only enabled the `std` feature of the crate for this crates `std` feature. This means `LittleEndian` and its `ByteOrder` trait are available on `no_std`, but `{ReadBytesExt, WriteBytesExt}` are not.
- Added a feature `embedded`, which enables two additional trait, which replace `std::io::{Read, Write}` (and the corresponding extensions from the `byteorder` crate) and implement the methods required.
- Changed the `write_x_message` methods to return results with a new error type `MessageWriteError` (analogous to `MessageReadError`). This was necessary, because `std::io::Error`s are obviously not available without `std`. For the read methods, I solved this by simply replacing the `Io` enum variant with one without an inner error (since the Error type for `embedded_hal::serial::{Read,Write}` is implementation-specific). This should be the only non-backwards-compatible change.
- Changed the methods in the connection trait accordingly.

I did not implement a connection for embedded serial ports yet. That connection would probably involve boxed traits for `Read<u8>` and `Write<u8>`.

Also, actually using this on an STM32F4 requires nightly rust because of the allocator, and to actually get it on the thing required compiling with size optimizations (`opt-level = "z"`).

<details>
  <summary>A working example on an STM32F411 ("black pill") with an ArduPilot connected:</summary>

```rust
#![no_main]
#![no_std]
#![feature(alloc_error_handler)]
#![feature(default_alloc_error_handler)]

use core::panic::PanicInfo;
use alloc_cortex_m::CortexMHeap;
use rtt_target::{rtt_init_print, rprintln};

use embedded_hal::serial::Write;
use mavlink::*;

use cortex_m;
use stm32f4xx_hal as hal;
use hal::{prelude::*, stm32};
use hal::serial::Serial;
use hal::serial::config::Config;

// Needed for alloc in rust-mavlink
#[global_allocator]
static ALLOCATOR: CortexMHeap = CortexMHeap::empty();

fn request_data_stream<W: Write<u8>>(
    target: &mut W,
    seq: &mut u8,
    stream: u8,
    freq: u16,
    startstop: u8
) -> Result<(), error::MessageWriteError> {
    let request = ardupilotmega::MavMessage::common({
        common::MavMessage::REQUEST_DATA_STREAM(
            common::REQUEST_DATA_STREAM_DATA {
                target_system: 1,
                target_component: 0,
                req_stream_id: stream,
                req_message_rate: freq,
                start_stop: startstop,
            },
        )
    });

    let header = MavHeader {
        system_id: 0xff,
        component_id: 0xbe,
        sequence: *seq,
    };

    *seq = seq.wrapping_add(1);

    write_versioned_msg::<ardupilotmega::MavMessage, _>(
        target,
        MavlinkVersion::V1,
        header,
        &request
    )
}

#[cortex_m_rt::entry]
fn main() -> ! {
    rtt_init_print!();
    rprintln!("RTT init");

    let dp = stm32::Peripherals::take().unwrap();
    let mut cp = cortex_m::peripheral::Peripherals::take().unwrap();

    cp.DWT.enable_cycle_counter();

    let gpioa = dp.GPIOA.split();

    // Configure clock to use 25MHz external oscillator and set PLL to 4x to
    // reach the maximum clock speed of 100MHz.
    let rcc = dp.RCC.constrain();
    let clocks = rcc.cfgr
        .use_hse(25.mhz())
        .sysclk(100.mhz())
        .freeze();

    // Initialize Heap
    let start = cortex_m_rt::heap_start() as usize;
    let size = 1024;
    unsafe { ALLOCATOR.init(start, size) }

    // Initialize MavLink serial port
    let tx = gpioa.pa9.into_alternate_af7();
    let rx = gpioa.pa10.into_alternate_af7();
    let serial = Serial::usart1(
        dp.USART1,
        (tx, rx),
        Config::default().baudrate(57600.bps()),
        clocks
    ).unwrap();
    let (mut mavlink_tx, mut mavlink_rx) = serial.split();

    let mut requested_data_streams = false;
    let mut seq_tx = 0;

    loop {
        match read_versioned_msg::<ardupilotmega::MavMessage, _>(&mut mavlink_rx, MavlinkVersion::V1) {
            Ok((_header, msg)) => {
                rprintln!("{:?}", msg);

                match msg {
                    ardupilotmega::MavMessage::common(common::MavMessage::HEARTBEAT(_data)) => {
                        if !requested_data_streams {
                            request_data_stream(&mut mavlink_tx, &mut seq_tx, 0, 0, 0); // disable all
                            request_data_stream(&mut mavlink_tx, &mut seq_tx, 10, 10, 1); // enable EXTRA1 (10 Hz)
                            requested_data_streams = true;
                        }
                    },
                    ardupilotmega::MavMessage::common(common::MavMessage::ATTITUDE(data)) => {
                        rprintln!("{:?}", data.pitch);
                    },
                    _ => {
                    }
                }
            },
            Err(e) => {
                rprintln!("error: {:?}", e);
            }
        }
    }
}

#[inline(never)]
#[panic_handler]
fn panic(info: &PanicInfo) -> ! {
    rprintln!("{}", info);
    loop {} // You might need a compiler fence in here.
}

```
</details>

I'm not sure if this is ready to be merged in its current state, but I figured I'd create a PR for input and discussion. I have also not been able to properly test the non-embedded version with these changes, but the example still compiles.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Check Type
       run: cargo fmt -- --check
     - name: Run internal tests
-      run: cargo test --verbose --all-features -- --nocapture
+      run: cargo test --verbose --features all-dialects -- --nocapture
 
   build:
     needs: quick-tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,18 +26,28 @@ jobs:
         include:
         - os: macos-latest
           TARGET: x86_64-apple-darwin
+          FEATURES: --features all-dialects
 
         - os: ubuntu-latest
           TARGET: arm-unknown-linux-musleabihf
+          FLAGS: --features ardupilotmega
 
         - os: ubuntu-latest
           TARGET: armv7-unknown-linux-musleabihf
+          FLAGS: --features ardupilotmega
 
         - os: ubuntu-latest
           TARGET: x86_64-unknown-linux-musl
+          FLAGS: --features all-dialects
+
+        - os: ubuntu-latest
+          TARGET: thumbv7m-none-eabi
+          FLAGS: --no-default-features --features embedded
 
         - os: windows-latest
           TARGET: x86_64-pc-windows-msvc
+          FLAGS: --features all-dialects
+
     steps:
     - name: Building ${{ matrix.TARGET }}
       run: echo "${{ matrix.TARGET }}"
@@ -51,4 +61,4 @@ jobs:
       with:
         use-cross: true
         command: build
-        args: --verbose --release --target=${{ matrix.TARGET }}
+        args: --verbose --release --target=${{ matrix.TARGET }} ${{ matrix.FLAGS }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2018"
 
 [build-dependencies]
 crc-any = "2.3.0"
-bytes = "0.4"
+bytes = { version = "1.0", default-features = false }
 xml-rs = "0.2"
 quote = "0.3"
 lazy_static = "1.2.0"
@@ -24,13 +24,15 @@ required-features = ["ardupilotmega"]
 
 [dependencies]
 crc-any = "2.3.5"
-bytes = "0.4"
-num-traits = "0.2"
+bytes = { version = "1.0", default-features = false }
+num-traits = { version = "0.2", default-features = false }
 num-derive = "0.3.2"
 bitflags = "1.2.1"
 serial = { version = "0.4", optional = true }
 serde = { version = "1.0.115", optional = true }
-byteorder = { version = "1.3.4", optional = true }
+byteorder = { version = "1.3.4", default-features = false }
+embedded-hal = { version = "0.2", optional = true }
+nb = { version = "0.1", optional = true }
 
 [features]
 "ardupilotmega" = ["common", "icarous", "uavionix"]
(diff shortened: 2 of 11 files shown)


KoffeinFlummi:
Looks like the CI fails (apart from some rustfmt) because it compiles with `--allfeatures`, which doesn't work because the `std` and `embedded` features cannot be used together. I'm not sure if there is a better solution for this than just listing all the dialect features in the build script?

patrickelectric:
Hi @KoffeinFlummi, sorry for taking so long to check your PR.
I'm improving our test environment and CI, and after finishing it I'll take a look in your PR to merge it with the proper feature flags.

patrickelectric (commented):
Thanks for the contribution!
Please be free to add any examples on our README or project related to embedded projects

KoffeinFlummi:
Thank you for merging! Once I finally get around to finishing the project I needed this for, I'll try to remember to make a PR to add it to the project list.

patrickelectric:
Awesome! I hope that everything goes well.
I have create a PR #94 that adds the embedded example, be free to check it out and review

KoffeinFlummi:
Looks great!

PR #87, opened 2021-03-07
Fix rust analyzer

patrickelectric:
A couple of developers including myself finds it much harder to code with rust-mavlink without autocomplete provided by the official tool rust-analyzer.
This patches avoid a problem where the generated files / includes depends of cargo environment variables to be find, since cargo does not provide such variables, since they are only available during compilation time, this patch moves the generated files under a know folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+src/generated
 target
 Cargo.lock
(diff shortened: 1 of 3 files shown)


kevinmehall:
The better way is to set the rust-analyzer option `loadOutDirsFromCheck`, soon to be renamed to `runBuildScripts`. `CARGO_MANIFEST_DIR` is not supposed to be writable (it will be in the cargo cache if this is installed as a dependency of another crate)

patrickelectric:
Thanks @kevinmehall I'll revert the PR, I'm talking with the rust-analyzer devs to find what is exactly wrong

PR #94, opened 2021-03-10
Add embedded example with test

patrickelectric:


podhrmic (approved):
I love the embedded example! Also it is super cool to see how is rust for embedded devices progressing:-)

PR #102, opened 2021-09-28
Moves rust-mavlink to move from serial to serialport

patrickelectric:
The main idea of this patch is to allow read and write in parallel when using serial port, as UDP and TCP methods already provide, serialport library has a full duplex implementation of the serial access to allow such thing.


podhrmic (approved):
LGTM

tstellanova (approved):

PR #107, opened 2021-10-21
bump 'xml-rs' dependency

danieleades:


Erk-:
This have been superseded by #123, so I think it can be closed

patrickelectric:
Thanks for the PR @danieleades

Issue #112, opened 2022-07-20
Commands are not being properly parsed when the command type originates from `ardupilotmega.xml` (and others)


bitsauce:
Hi! I have been trying to receive a command that is defined in `ardupilotmega.xml` without any success. I found out that this doesn't currently work because the generated `mavlink::common::COMMAND_LONG` struct has a member `pub command: mavlink::common::MavCmd`, however, that enum currently does not contain commands from any other file than `common.xml`.

This results in `mavlink::common::COMMAND_LONG_DATA::deser()` failing here:

```rust
_struct.command = FromPrimitive::from_u16(tmp).ok_or(ParserError::InvalidEnum {
    enum_type: "MavCmd".to_string(),
    value: tmp as u32,
})?;
```

Any thoughts as to how this should be solved?

Issue #59, opened 2020-06-19
ardupilotmega is not using icarous and uavionix modules when generated


patrickelectric:
@laptou can you take a look ?

patrickelectric:
It would be nice to have access to common and others messages that are included in ardupilotmega with:
```
pub fn get_apm_mount_status() -> mavlink::ardupilotmega::HEARTBEAT {
    mavlink::ardupilotmega::HEARTBEAT {
        custom_mode: 0,
        mavtype: mavlink::ardupilotmega::MavType::MAV_TYPE_QUADROTOR,
        autopilot: mavlink::ardupilotmega::MavAutopilot::MAV_AUTOPILOT_ARDUPILOTMEGA,
        base_mode: mavlink::ardupilotmega::MavModeFlag::empty(),
        system_status: mavlink::ardupilotmega::MavState::MAV_STATE_STANDBY,
        mavlink_version: 0x3,
    }
}
```

patrickelectric:
Maybe is better to check the necessary includes in the xml and merge all in a big one to do the parse process.

PR #113, opened 2022-07-22
Changed mavlink definitions parsing such that messages and enums from included files are included in the generated output file and flattened

bitsauce:
As mentioned in issue #112, any extended commands that originate from a dialect file other than `common.xml` will not be parsed properly.

The solution I implemented in this PR is to simply flatten all definitions (enums and messages) from every included xml file. This approach is also mentioned in #59. As is noted in that issue, this does change the API slightly in that all the definitions that are supported within a dialect (including definitions that come from included files) will now be available under the dialect namespace itself (e.g. we have to use `mavlink::ardupilotmega::MavMessage::HEARTBEAT` instead of `mavlink::common::MavMessage::HEARTBEAT` if we're using `ardupilotmega`)

Let me know what you think


patrickelectric:
@bitsauce please check the CI.

bitsauce:
Seems like some change in `master` broke my implementation. I'll get back to you if I manage to fix it

bitsauce:
Seems like my assessment was incorrect; what's going on here is that the CI runner is running out of memory when it's trying to compile the test programs. This is probably happening in this PR, because when you flatten out all the definitions, it means a lot of definitions are duplicated and must be kept in memory, especially when compiling with `--features all-dialects`. `common.rs` is, by itself, 14K lines, which means that when `--features all-dialects` is used, you will have to compile the same 14K lines for each dialect that includes common.

I'm not sure what the ideal solution is. One solution would be to split up `cargo test` such that it is called once for every dialect (so that all dialects won't have to be compiled and stored in memory at once). Maybe there exists some way to prevent duplicate definitions, while simultaneously keeping the messages flattened within the namespaces, not sure.

patrickelectric:
@bitsauce that makes sense, be free to break the test case, it makes sense from what I can see around your approach.

patrickelectric:
@bitsauce I should check out the PR friday, ping me if there is no review from me until monday

bitsauce:
Did you have a chance to review this PR @patrickelectric ?

patrickelectric (commented):
I'm doing some tests with the code to see the final result today, will give further feedback before the end of the day

patrickelectric (changes requested):
The functionality is correct, I would just recommend to merge the commits in a single commit or break it in a more simpler logic, the amount of commits is unnecessary since they are mostly fixing changes related to previous commits of the same patch.
A couple of commits such as:
- Update library to use flat pattern
- Update examples to use new version of library
- Update CI to build each dialect individually

I would also recommend to move ardupilotmega to be the default dialect

patrickelectric (approved):

patrickelectric:
Thanks @bitsauce!

PR #115, opened 2022-08-11
Add read_*_raw_message

patrickelectric:
Allow checking if there is a valid message without parsing it


joaoantoniocardoso (approved):
Looks great, it will help applications where the specificities of the messages are not relevant.

PR #122, opened 2022-08-18
Truncate MAVLink 2 messages

Erk-:
The MAVLink documentation states the following[^1]:
> *MAVLink 2* implementations *must* truncate any empty (zero-filled)
> bytes at the end of the serialized payload before it is sent.

So this PR implements that.

[^1]: https://mavlink.io/en/guide/serialization.html#payload_truncation

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,6 +42,10 @@ mod connection;
 #[cfg(feature = "std")]
 pub use self::connection::{connect, MavConnection};
 
+mod utils;
+#[allow(unused_imports)]
+use utils::remove_trailing_zeroes;
+
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
@@ -71,7 +75,7 @@ where
 {
     fn message_id(&self) -> u32;
     fn message_name(&self) -> &'static str;
-    fn ser(&self) -> Vec<u8>;
+    fn ser(&self, version: MavlinkVersion) -> Vec<u8>;
 
     fn parse(
         version: MavlinkVersion,
@@ -159,9 +163,8 @@ impl<M: Message> MavFrame<M> {
                 v.push(self.msg.message_id() as u8); //TODO check
             }
         }
-
         // serialize message
-        v.append(&mut self.msg.ser());
+        v.append(&mut self.msg.ser(self.protocol_version));
 
         v
     }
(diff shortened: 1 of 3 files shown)


patrickelectric (commented):
Please squash the commits, I'll do some tests to validate the PR just to make sure.

Erk-:
I have rebased and squashed it, but it will likely need some minor changes when #124 is merged similar to the ones I have [here](https://github.com/fusion-engineering-forks/rust-mavlink/blob/fusion/mavlink-changes/build/util.rs#L22)

src/lib.rs
```diff
@@ -71,7 +71,8 @@ where
 {
     fn message_id(&self) -> u32;
     fn message_name(&self) -> &'static str;
-    fn ser(&self) -> Vec<u8>;
+    fn ser_v1(&self) -> Vec<u8>;
```
patrickelectric: Why not creating a `ser` function that receives as argument `MavlinkVersion` ? Like what we have for `deser`. The only difference of `ser_v1` and `ser_v2` is the trailing removal logic.

build/util.rs
```diff
@@ -9,3 +12,18 @@ pub fn to_module_name<P: Into<PathBuf>>(file_name: P) -> String {
...
+
+/// Removes the trailing zeroes in the payload
+///
+/// # Note:
+///
+/// There must always be at least one remaining byte even if it is a
+/// zero byte.
+pub fn q_remove_trailing_zeroes(buf: proc_macro2::Ident) -> TokenStream {
+    quote! {
+        while let Some(&0) = #buf.last() {
```
patrickelectric: Doing so we end up duplicating the same code again and again inside the binary, it's better to have a global function that abstract such behavior over having the same logic around every serialized message.

patrickelectric (changes requested):

build/util.rs
patrickelectric: To do that, we probably need to move this function to a library that will be used by all dialects.

build/util.rs
```diff
@@ -9,3 +12,18 @@ pub fn to_module_name<P: Into<PathBuf>>(file_name: P) -> String {
...
         .replace(|c: char| !c.is_alphanumeric(), "_") // remove non alphanum
 }
+
+/// Removes the trailing zeroes in the payload
+///
+/// # Note:
+///
+/// There must always be at least one remaining byte even if it is a
+/// zero byte.
+pub fn q_remove_trailing_zeroes(buf: proc_macro2::Ident) -> TokenStream {
```
patrickelectric: What is the reason behind the q prefix ? That sounds unnecessary.

patrickelectric (approved):

patrickelectric (approved):
Thanks!

PR #123, opened 2022-08-18
Use quick-xml instead of xml-rs

Erk-:
Since XML-rs is for the time being unmaintained[^1], and have been
marked as such in RUSTSEC[^2]. I have ported the XML parsing to
quick-xml.

Supersedes #107 

[^1]: https://github.com/netvl/xml-rs/issues/221

[^2]: https://rustsec.org/advisories/RUSTSEC-2022-0048

### DIFF
--- a/build/parser.rs
+++ b/build/parser.rs
@@ -1,10 +1,11 @@
 use crc_any::CRCu16;
 use std::cmp::Ordering;
 use std::default::Default;
-use std::io::{Read, Write};
+use std::io::{BufRead, BufReader, Read, Write};
 use std::u32;
 
-use xml::reader::{EventReader, XmlEvent};
+//use xml::reader::{EventReader, XmlEvent};
+use quick_xml::{events::Event, Reader};
 
 use quote::{Ident, Tokens};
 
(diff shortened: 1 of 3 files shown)


build/parser.rs
```diff
@@ -1036,24 +1037,24 @@ pub enum MavXmlElement {
...
-        "entry" => Some(Entry),
-        "description" => Some(Description),
-        "param" => Some(Param),
-        "messages" => Some(Messages),
-        "message" => Some(Message),
-        "field" => Some(Field),
-        "deprecated" => Some(Deprecated),
-        "wip" => Some(Wip),
-        "extensions" => Some(Extensions),
+        b"version" => Some(Version),
```
patrickelectric, line 1043: Isn't it better to transform the xml access to utf-8 over than using all our strings to bytes ?

Erk-: Do you mean to convert it to strings when we compare, or do you mean to use the Decoder of quick-xml/unescape_and_decode?

quick-xml does all access by default in bytes, i tried to only convert it to strings where it is needed.

patrickelectric: Ok, I think that we continue with the current version of the PR.

patrickelectric (changes requested):
CI is failing, I would recommend to check and test the CI in your own fork. Right now I need to approve the CI in the repository everytime that you update the PR.

Erk-:
> CI is failing, I would recommend to check and test the CI in your own fork. Right now I need to approve the CI in the repository everytime that you update the PR.

Seems it was only the fmt, it looks to get through CI now at least https://github.com/fusion-engineering-forks/rust-mavlink/runs/7946114295

PR #132, opened 2022-09-05
Update mavlink submodule

Erk-:
This patch updates the mavlink submodule and ads some fixes to make it work with the new definitions. 

I added a patch because the arrays of bit masks are quite annoying to support with the current way they are generated, I have asked what they think about it upstream or else we will just have to change it up. https://github.com/mavlink/mavlink/issues/1884 

The changes in #113 will probably make the changes in [src/bin/mavlink-dump.rs](https://github.com/mavlink/rust-mavlink/compare/master...fusion-engineering-forks:rust-mavlink:valdemar/update-mavlink-definitions#diff-82c9e68044b14e96c9dc03f4fa3d3d7da7c20e6bd2981c5b05e65bbdd49ca1b9) unnecessary if that is merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ nb = { version = "1.0", optional = true }
 "ualberta" = ["common"]
 "uavionix" = ["common"]
 "icarous" = []
-"common" = []
+"common" = ["minimal"]
 
 "all-dialects" = [
     "ardupilotmega",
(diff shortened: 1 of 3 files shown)


Erk-:
Found some more issues when I added the new dialects, I think they will be mostly solved by 113 so I will put this a draft until there is an update there

Issue #139, opened 2022-11-25
MavFrame serialize into a vector with wrong fields order


qwerty19106:
True fields order for [Mavlink Serialization](https://mavlink.io/en/guide/serialization.html):
- magic (start)
- len
- ... (V2 fields)
- sequence
- system_id
- component_id
- message_id
- payload
- checksum

Now (**MavFrame::ser** method):
- system_id
- component_id
- sequence
- message_id
- payload

In addition, **MavFrame::ser** and **MavFrame::deser** methods are not used in the **MavConnection** implementors as expected.

PR #140, opened 2022-11-26
[WIP] Message serialization refactoring

qwerty19106:
This refactoring optimizes code for embedded context and tiny increases performance generally.
Also it simplify **AsyncReader** code, which I plan to send in next PR.

1) Message serialization moved to **MAVLinkV1MessageRaw::serialize_message** and **MAVLinkV2MessageRaw::serialize_message** methods to get common serialization API. The **write_v1_msg** and **write_v2_msg** methods has been left as thin wrapper.
As variant, this code should be placed to **MavFrame::ser** method. But **MavFrame::ser** body causes the some [questions](https://github.com/mavlink/rust-mavlink/issues/139).

2) **MAVLinkV1MessageRaw** and **MAVLinkV2MessageRaw** has been make as thin struct over byte array. It allows send serialized message over [Direct Memory Acces](https://en.wikipedia.org/wiki/Direct_memory_access) on **embedded** and other contexts. DMA is very fast in comparison with interrupts or polling.
It break **MAVLinkV1MessageRaw** and **MAVLinkV2MessageRaw** API: fields becomes methods.

3) I removed unnecessary bytes copying at payload serialization. **BytesMut** was been rewrite as thin wrapper over slice.
It is breaking change: **Message::ser** method was changed. But I hope that no one uses **Message::ser** method directly.


qwerty19106:
The v1_encode_decode_tests is failed, but CI is succesfull. It is strange.

qwerty19106:
It is ready to review

src/bytes_mut.rs
```diff
@@ -133,11 +124,11 @@ impl<const N: usize> BytesMut<N> {
...
-    #[inline]
-    fn deref(&self) -> &[u8] {
-        &self.data[..self.len]
-    }
-}
+//     #[inline]
+//     fn deref(&self) -> &[u8] {
+//         &self.data[..self.len]
+//     }
+// }
```
patrickelectric: You can remove code that is not necessary

patrickelectric (changes requested):

patrickelectric:
The code looks great, the only thing necessary is to remove the old commented code.

src/lib.rs
```diff
@@ -209,45 +213,105 @@ pub fn read_versioned_msg<M: Message, R: Read>(
...
-    pub checksum: u16,
-}
+pub struct MAVLinkV1MessageRaw([u8; 1 + Self::HEADER_SIZE + 255 + 2]);
 
 impl MAVLinkV1MessageRaw {
+    const HEADER_SIZE: usize = 5;
+
+    pub const fn new() -> Self {
+        Self([0; 1 + Self::HEADER_SIZE + 255 + 2])
+    }
```
danieleades, line 228: worth a `Default` implementation?

src/lib.rs
```diff
@@ -209,45 +213,105 @@ pub fn read_versioned_msg<M: Message, R: Read>(
...
+
+    #[inline]
+    pub fn header(&mut self) -> &[u8] {
+        &self.0[1..(1 + Self::HEADER_SIZE)]
```
danieleades: is this equivalent to

```rust
&self.0[1..=Self::HEADER_SIZE]
```
?

src/lib.rs
```diff
@@ -315,49 +369,129 @@ const MAVLINK_IFLAG_SIGNED: u8 = 0x01;
...
-}
+pub struct MAVLinkV2MessageRaw([u8; 1 + Self::HEADER_SIZE + 255 + 2 + Self::SIGNATURE_SIZE]);
 
 impl MAVLinkV2MessageRaw {
+    const HEADER_SIZE: usize = 9;
+    const SIGNATURE_SIZE: usize = 13;
+
+    pub const fn new() -> Self {
+        Self([0; 1 + Self::HEADER_SIZE + 255 + 2 + Self::SIGNATURE_SIZE])
+    }
```
danieleades, line 391: impl `Default`

src/lib.rs
qwerty19106: Yes, I added Default.

src/lib.rs
qwerty19106: Yes.
But other methods (such as MAVLinkV1MessageRaw::payload) will become confused if I remove "1 + ".
For example MAVLinkV1MessageRaw::payload:
```&self.0[(1 + Self::HEADER_SIZE)..(1 + Self::HEADER_SIZE + payload_length)]```
to
```&self.0[(1 + Self::HEADER_SIZE)..=(Self::HEADER_SIZE + payload_length)]```.

Are you sure this change is necessary?

Comment (danieleades) [reply src/lib.rs]: Only where it makes the intent clearer. Where it *doesn't*, I would consider checking whether clippy complains (enable 'pedantic and 'nursery' lints for this) and explicitly suppress the lint for that line

File: src/lib.rs [hunk 5]
```diff
@@ -209,45 +213,111 @@ pub fn read_versioned_msg<M: Message, R: Read>(
...
+impl Default for MAVLinkV1MessageRaw {
+    fn default() -> Self {
+        Self::new()
+    }
 }
 
 impl MAVLinkV1MessageRaw {
+    const HEADER_SIZE: usize = 5;
+
+    pub const fn new() -> Self {
```
patrickelectric, line 226: The `new` function is usually used when there are arguments to construct the object. The functionality should move to `default` and `new`could be removed.

qwerty19106: The ```new ``` is const fn, but I could not make ```impl const Default``` in stable Rust. It requires ```const_trait_impl``` feature.

patrickelectric: Ok!

patrickelectric (approved):

patrickelectric:
I would just recommend to squash the latest commit on the specific commits that it's fixing.
@danieleades, any more comments ?

src/lib.rs
```diff
@@ -263,21 +332,11 @@ pub fn read_v1_raw_message<R: Read>(
...
-        system_id: reader.read_u8()?,
-        component_id: reader.read_u8()?,
-        message_id: reader.read_u8()?,
-        payload_buffer: [0; 255],
-        checksum: 0,
-    };
-
-    let payload = &mut message.payload_buffer[..message.payload_length as usize];
-    reader.read_exact(payload)?;
+    let mut message = MAVLinkV1MessageRaw::new();
```
danieleades, line 335:
```suggestion
    let mut message = Self::new();
```

src/lib.rs
```diff
@@ -315,49 +374,135 @@ const MAVLINK_IFLAG_SIGNED: u8 = 0x01;
...
+
+    #[inline]
+    pub fn message_id(&self) -> u32 {
+        u32::from_le_bytes([self.0[7], self.0[8], self.0[9], 0])
+    }
+
+    #[inline]
     pub fn payload(&self) -> &[u8] {
-        &self.payload_buffer[..self.payload_length as usize]
+        let payload_length = self.payload_length() as usize;
```
danieleades: lossless conversions-

```suggestion
        let payload_length: u8 = self.payload_length().into();
```

the type annotation might not even be required here

src/lib.rs
```diff
@@ -315,49 +374,135 @@ const MAVLINK_IFLAG_SIGNED: u8 = 0x01;
...
+        &self.0[(1 + Self::HEADER_SIZE)..(1 + Self::HEADER_SIZE + payload_length)]
     }
 
-    pub fn mut_payload(&mut self) -> &mut [u8] {
-        &mut self.payload_buffer[..self.payload_length as usize]
+    #[inline]
+    pub fn checksum(&self) -> u16 {
+        let payload_length = self.payload_length() as usize;
```
danieleades: use lossless conversion

src/lib.rs
```diff
@@ -315,49 +374,135 @@ const MAVLINK_IFLAG_SIGNED: u8 = 0x01;
...
+        u16::from_le_bytes([
+            self.0[1 + Self::HEADER_SIZE + payload_length],
+            self.0[1 + Self::HEADER_SIZE + payload_length + 1],
+        ])
+    }
+
+    fn mut_payload_and_checksum_and_sign(&mut self) -> &mut [u8] {
+        let payload_length = self.payload_length() as usize;
```
danieleades: use lossless conversion

i won't bother writing all of these out. You should be find them all with something like

`cargo +nightly clippy -- -A clippy::all -W clippy::cast_lossless`

src/utils.rs
```diff
@@ -1,14 +1,11 @@
-use crate::{bytes_mut::BytesMut, MAX_FRAME_SIZE};
-
 /// Removes the trailing zeroes in the payload
 ///
 /// # Note:
 ///
 /// There must always be at least one remaining byte even if it is a
 /// zero byte.
 #[allow(dead_code)]
```
danieleades: why do we need to allow dead code here?

danieleades (changes requested):
Just a couple of little nits

src/lib.rs
qwerty19106: The `read_v1_raw_message` function is not method. I could not apply `Self` here.

src/utils.rs
patrickelectric: That's not related to the PR.

src/lib.rs
qwerty19106: The `payload_length` method return `u8`. Casting from u8 to usize is [loseless](https://doc.rust-lang.org/reference/expressions/operator-expr.html#type-cast-expressions).
> Casting from a smaller integer to a larger integer (e.g. u8 -> u32) will
>   zero-extend if the source is unsigned

I could change `payload_length` return type to `usize` if needed.

qwerty19106:
@patrickelectric, could I squash all commits to single commit?

src/lib.rs
danieleades: > The `payload_length` method return `u8`. Casting from u8 to usize is [loseless](https://doc.rust-lang.org/reference/expressions/operator-expr.html#type-cast-expressions).
> > Casting from a smaller integer to a larger integer (e.g. u8 -> u32) will
> >   zero-extend if the source is unsigned
> 
> I could change `payload_length` return type to `usize` if needed.

I understand the cast is lossless. That's why you might as well use the *explicitly* lossless method.

Doing the conversion once, inside the `payload_length` method, sounds like a good approach

src/utils.rs
qwerty19106: It was `#[allow(dead_code)] ` before this PR. I will remove it.

src/lib.rs
danieleades: 👍

patrickelectric:
> @patrickelectric, could I squash all commits to single commit?

Yes.

src/lib.rs
patrickelectric: the mavlink protocol defines the payload_length as `u8`, we should stay with it to make the protocol definition clear.

patrickelectric:
Thanks for the PR @qwerty19106, nice work!

PR #142, opened 2022-11-29
add msrv target to CI

### LINKED ISSUE #137: add MSRV
this project doesn't seem to define an MSRV.

This should be added to CI.

benefits-

- bumping the MSRV should be considered a patch change pre-1.0, and a minor change post 1.0
- clippy can be configured with the MSRV to stop it suggesting changes which the MSRV doesn't support

the current MSRV for this project looks to be 1.56.1

danieleades:
closes #137 

question: what 'features' should be enabled when checking the MSRV? what will give the greatest coverage of the codebase?

question: where in the CI should this live? right now it's a separate job, but rustfmt and clippy are inside the 'quick-checks' job


.github/workflows/test.yml
```diff
@@ -33,6 +34,23 @@ jobs:
...
+  msrv:
+    needs: quick-tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+    - uses: actions/checkout@master
+    - uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: "1.60.0"
```
patrickelectric, line 46: Doesn't it make more sense to move the `build` step to use 1.60 ?

danieleades: Not necessarily. The usual approach is to run your tests and everything on stable (or nightly), and just run a check on your MSRV to check you haven't added dependencies or syntax that force downstream users to bump their compiler version.
But your primary target is still the newer compiler, and you want to opt in to any compiler warnings or errors that might not be present in the older compiler.

This is how I usually set this up- https://github.com/danieleades/monzo-lib/blob/main/.github/workflows/CI.yml (this is a pretty standard approach)

danieleades: Testing the beta compiler is definitely overkill...

Maybe a better approach is to add an MSRV target to the main build matrix?

patrickelectric: Thanks for the explanation!

patrickelectric (approved):

PR #143, opened 2022-11-29
update clippy linting job

danieleades:
the old action is unsupported. This one will add inline comments to PRs. pretty.

question - what feature set should be enabled when performing clippy linting?

i'd prefer for the clippy warnings not to be escalated to build failures, so I'll leave this as a draft until I have a resolution there - see https://github.com/giraffate/clippy-action/issues/23

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,13 +20,14 @@ jobs:
       - uses: actions/checkout@master
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2022-11-30
+          toolchain: nightly-2022-12-01
           components: clippy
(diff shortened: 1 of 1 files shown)


patrickelectric (changes requested):
Thanks for the PR, please merge the development commits for us to merge the code.

danieleades:
> Thanks for the PR, please merge the development commits for us to merge the code.

I don't think this one's ready for the big time yet. See https://github.com/giraffate/clippy-action/issues/23

danieleades:
upstream github action has been fixed. Just waiting on a release, and then this will be ready for review

danieleades:
i take it back. the action is still escalating clippy warnings to hard build failures, just in a different way...

PR #152, opened 2023-02-15
fix: correctly deserialize MavFrame

GabrielDertoni:
This commit addresses two bugs present in the `MavFrame::deser` implementation. It parsed the sequence number after system and component ids, but should instead parse it before. Furthermore the implementation used the `get_u32_le` method to parse the message id in mavlink v2. However the MAVLink spec specifies that this field should have 3 bytes, not 4 (like `u32`).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -172,9 +172,9 @@ impl<M: Message> MavFrame<M> {
     pub fn deser(version: MavlinkVersion, input: &[u8]) -> Result<Self, ParserError> {
         let mut buf = Bytes::new(input);
 
+        let sequence = buf.get_u8();
         let system_id = buf.get_u8();
         let component_id = buf.get_u8();
-        let sequence = buf.get_u8();
         let header = MavHeader {
             system_id,
             component_id,
@@ -182,8 +182,8 @@ impl<M: Message> MavFrame<M> {
         };
 
         let msg_id = match version {
-            MavlinkVersion::V2 => buf.get_u32_le(),
-            MavlinkVersion::V1 => buf.get_u8().into(),
+            MavlinkVersion::V2 => buf.get_u24_le(),
+            MavlinkVersion::V1 => buf.get_u8() as u32,
         };
 
         match M::parse(version, msg_id, buf.remaining_bytes()) {
--- a/tests/mav_frame_tests.rs
+++ b/tests/mav_frame_tests.rs
@@ -1,0 +1,34 @@
+pub mod test_shared;
+
+mod mav_frame_tests {
+    // NOTE: No header
+    pub const HEARTBEAT_V2: &[u8] = &[
+        0xef, // seq 239
+        0x01, // sys ID
+        0x01, // comp ID
+        0x00, 0x00, 0x00, // msg ID
+        0x05, 0x00, 0x00, 0x00, 0x02, 0x03, 0x59, 0x03, 0x03, // payload
+        16, 240, // checksum
+    ];
+
+    #[test]
+    pub fn test_deser() {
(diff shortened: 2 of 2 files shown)


patrickelectric (approved):

patrickelectric:
@GabrielDertoni please check the CI

patrickelectric:
@GabrielDertoni since your last commits are just corrections of the first one, can you squash everything in the same commit ?

GabrielDertoni:
Done, sorry I though the squash would be done automatically on merge

patrickelectric:
We only have rebase squash for cleaner git history

patrickelectric:
@GabrielDertoni check rustfmt

patrickelectric:
ping @GabrielDertoni

GabrielDertoni:
Just noticed the same bug in `MavFrame::ser`. Should I open a new PR or can I add the fix here as well?

patrickelectric:
> Just noticed the same bug in `MavFrame::ser`. Should I open a new PR or can I add the fix here as well?

Is this being introduced in this PR ?

src/lib.rs
```diff
@@ -170,18 +170,22 @@ impl<M: Message> MavFrame<M> {
...
         };
 
         let msg_id = match version {
-            MavlinkVersion::V2 => buf.get_u32_le(),
-            MavlinkVersion::V1 => buf.get_u8().into(),
+            MavlinkVersion::V2 => {
+                let mut bytes = [0u8; 4];
+                bytes[..3].copy_from_slice(buf.get_bytes(3));
+                u32::from_le_bytes(bytes)
+            }
```
Erk-: Would it not make more sense to add a `get_u24_le` method to our bytes implementation instead of handling it in a special way here.

qwerty19106:
I don't understand what functionality the `MavFrame` implements?
1) The`MavFrame::deser` method could not be used because we should found start and end of frame.
The `MavConnection::recv` methods calls `read_versioned_msg`, because there is no reliable way found start and end of frame without parsing of header.
It is possible on microcontrollers, for example IDLE event of UART peripheral. But it is more complicated theme.
2) The`MavFrame::ser` method uses `header` field, but `MavHeader` should be reused because `system_id` and `component_id` is constant for a specific device.
For this reason the `MavConnection::send` method consumes `&MavHeader`.

Thus I propose remove `MavFrame` fully as unnecessary and confusing API.
@Erk-

src/lib.rs
Erk-: Something like the following should work:
```rust
    pub fn get_u24_le(&mut self) -> u32 {
	const SIZE: usize = 3;
	self.check_remaining(SIZE);

	let mut val = [0u8; SIZE + 1];
	val[..3].copy_from_slice(&self.data[self.pos..self.pos + SIZE]);	
        self.pos += SIZE;
        u32::from_le_bytes(val)
    }

    pub fn get_i24_le(&mut self) -> i32 {
	const SIZE: usize = 3;
	self.check_remaining(SIZE);

	let mut val = [0u8; SIZE + 1];
	val[..3].copy_from_slice(&self.data[self.pos..self.pos + SIZE]);	
        self.pos += SIZE;
        i32::from_le_bytes(val)
    }
```

GabrielDertoni:
I actually use `MavFrame` as just a way to store header, message and version info in the same place. Even though I have spotted the bug in `ser` and `deser` implementations, I haven't really used those since the API is not great.

src/lib.rs
Erk-: This have been implemented in #159 so you can use that

patrickelectric:
@GabrielDertoni can you rebase and use #159 new features ?

patrickelectric (approved):

Issue #156, opened 2023-03-19
Discussion: A better `Message` trait


GabrielDertoni:
## Introduction

Currently the `Message` trait is defined like this

```rust
pub trait Message
where
    Self: Sized,
{
    fn message_id(&self) -> u32;
    fn message_name(&self) -> &'static str;

    /// Serialize **Message** into byte slice and return count of bytes written
    fn ser(&self, version: MavlinkVersion, bytes: &mut [u8]) -> usize;

    fn parse(
        version: MavlinkVersion,
        msgid: u32,
        payload: &[u8],
    ) -> Result<Self, error::ParserError>;

    fn message_id_from_name(name: &str) -> Result<u32, &'static str>;
    fn default_message_from_id(id: u32) -> Result<Self, &'static str>;
    fn extra_crc(id: u32) -> u8;
}
```

Then types implementing `Message` are structs of several `*_DATA` types that store the actual data for the message. Each of those types has `ser` and `deser` methods but no unifying trait to encompass this behaviour. They also have an associated constant `ENCODED_LEN` and implement `Default`.

## Current limitations and issues

1. Returning a `Result<T, &'static str>` isn't very helpfull in this case. There is a single reason for this to fail: the massage with the given name/id doesn't exist! However, the type indicates that many different error messages could be emitted. A better type would be `Result<T, MessageDoesNotExist>` or simply `Option<T>`.
2. `message_id_from_name`, `default_message_from_id`, `extra_crc`, `message_id` and `message_name` are all trying to get some metadata about the message, though some do that with an instance of the message and others are static. We could instead return a `&'static MessageMeta` type that stores all the metadata in a single place. This would allow stuff like getting the message name from an id without having to construct the message. We would also avoid having to do multiple calls if we want multiple pieces of metadata. The metadata struct can also be passed around which is pretty usefull.
3. In order to get a default instance of a message we need to construct it, even if we only need a reference to it. So, this `MessageMeta` struct could also store the default value for the message. Then we would have one `MessageMeta` static for every message `*_DATA`.
4. Currently there is a mandatory dependency on [`heapless`](https://crates.io/crates/heapless), but very little of the library is actually used (only `Vec`) and those uses could just be standard Rust arrays.
5. Instead of implementing `Bytes` and `BytesMut`, use the [`bytes`](https://crates.io/crates/bytes) crate. It's a tiny crate that does pretty much the same as this crate's implementation, but uses a trait instead of a type, and implements the trait for slices.

[...]

PR #160, opened 2023-03-20
refactor: remove heapless dependency

GabrielDertoni:
This PR aims to remove the [`heapless`](https://crates.io/crates/heapless) dependency. Currently the implementation relies on `heapless::Vec` but doesn't have to. Removing this dependency reduces the ammount of code generated and the compile time. It also uses a `const DEFAULT` for default values of `*_DATA` types which can be used in `const fn`s.

Look at #156 for more context.

### DIFF
--- a/build/parser.rs
+++ b/build/parser.rs
@@ -178,8 +178,6 @@ impl MavProfile {
             use num_traits::ToPrimitive;
             #[allow(unused_imports)]
             use bitflags::bitflags;
-            #[allow(unused_imports)]
-            use heapless::Vec;
 
             use crate::{Message, error::*, bytes::Bytes, bytes_mut::BytesMut};
 
@@ -410,10 +408,15 @@ impl MavEnum {
         quote!(#name)
     }
 
+    fn emit_const_default(&self) -> TokenStream {
+        let default = format_ident!("{}", self.entries[0].name);
+        quote!(pub const DEFAULT: Self = Self::#default;)
+    }
+
     fn emit_rust(&self) -> TokenStream {
         let defs = self.emit_defs();
-        let default = format_ident!("{}", self.entries[0].name);
         let enum_name = self.emit_name();
+        let const_default = self.emit_const_default();
 
         #[cfg(feature = "emit-description")]
         let description = if let Some(description) = self.description.as_ref() {
@@ -450,15 +453,19 @@ impl MavEnum {
             };
         }
 
-        dbg!(quote! {
+        quote! {
             #enum_def
+
+            impl #enum_name {
+                #const_default
+            }
 
             impl Default for #enum_name {
                 fn default() -> Self {
-                    Self::#default
+                    Self::DEFAULT
                 }
             }
-        })
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


build/parser.rs
```diff
@@ -179,7 +179,6 @@ impl MavProfile {
             #[allow(unused_imports)]
             use bitflags::bitflags;
             #[allow(unused_imports)]
```
danieleades: do you need to remove this flag?

build/parser.rs
```diff
@@ -582,12 +597,34 @@ impl MavMessage {
...
+            }
+        }
+    }
+
+    fn emit_const_default(&self) -> TokenStream {
+        let initializers = self
+            .fields
+            .iter()
+            .map(|field| field.emit_default_initializer())
+            .collect::<TokenStream>();
```
danieleades: does this work without the `collect`?

build/parser.rs
```diff
@@ -806,23 +861,11 @@ impl MavType {
...
-                    // handle as a slice
-                    let r = t.rust_reader(&quote!(let val), buf);
-                    quote! {
-                        for idx in 0..#size {
-                            #r
-                            #val[idx] = val;
-                        }
+                let r = t.rust_reader(&quote!(let val), buf);
+                quote! {
+                    for idx in 0..#size {
```
danieleades:
```suggestion
                    for v in &mut val {
```

(or something similar)

build/parser.rs
```diff
@@ -733,6 +771,21 @@ impl MavField {
...
         }
     }
+
+    fn emit_default_initializer(&self) -> TokenStream {
+        let field = self.emit_name();
+        // FIXME: Is this actually expected behaviour??
+        if matches!(self.mavtype, MavType::Array(_, _)) {
+            let default_value = self.mavtype.emit_default_value();
+            quote!(#field: #default_value,)
+        } else if let Some(ref enumname) = self.enumtype {
```
danieleades:
```suggestion
        } else if let Some(enumname) = &self.enumtype {
```

can match ergonomics help here?

patrickelectric:
@GabrielDertoni check ci

qwerty19106:
You can replace [serde_arrays](https://github.com/Kromey/serde_arrays) dependency on [serde-big-array](https://github.com/est31/serde-big-array) to fix `the trait 'Default' is not implemented for '[u8; 200]'` problem.

patrickelectric (approved):

patrickelectric:
It appears fine by be, is there any further comments around this PR ?

qwerty19106:
It would be nice to add a link to the issue into the `RustDefault` description.
The best I found is [issue](https://users.rust-lang.org/t/issue-for-derives-for-arrays-greater-than-size-32/59055/3).

> Default is special, because we already have impl<T> Default for [T; 0] without the T: Default constraint. Rewrite it in const generics requires the const generics specialization, which is far from the stable yet.

GabrielDertoni:
ping @patrickelectric

patrickelectric (approved):

PR #161, opened 2023-03-20
perf: improve code generation performance

GabrielDertoni:
Use `BufWriter` and avoid formatting generated code for faster build times.

**BEFORE**
rust_mavlink_master_comptime

**AFTER**
rust_mavlink_better_comptime

This PR still doesn't address the main issue which is the compilation time of the `mavlink` crate itself. But for a 14 line change I don't see a reason not to do this. It makes code generation 3x faster.

PS: The measurements were taken with `cargo clean` followed by `cargo b --test encode_decode_tests --timings` on a Ryzen 5 5600X processor.

PR #162, opened 2023-03-21
parallelise the code-gen

danieleades:
inspired by @GabrielDertoni's work in #161, this PR provides a (much more modest) performance boost by parallelising the code-gen using rayon

Screenshot 2023-03-21 080654

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ quote = "1"
 proc-macro2 = "1.0.43"
 lazy_static = "1.2.0"
 serde = { version = "1.0.115", optional = true, features = ["derive"] }
+rayon = "1.7.0"
 
 [[bin]]
 name = "mavlink-dump"
--- a/build/main.rs
+++ b/build/main.rs
@@ -4,8 +4,11 @@ mod binder;
 mod parser;
 mod util;
 
+use rayon::prelude::{ParallelBridge, ParallelIterator};
+
 use crate::util::to_module_name;
 use std::env;
+#[cfg(feature = "format-generated-code")]
 use std::ffi::OsStr;
 use std::fs::{read_dir, File};
 use std::io::BufWriter;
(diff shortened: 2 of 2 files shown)


danieleades:
- does rayon provide a speed up on all targets, or only some?
- is it worth feature-gating this?

GabrielDertoni:
I would like to see how this does after #161 gets merged. Maybe the 1.84s of rayon comp time will already make it not worth it.

Cargo.toml
```diff
@@ -51,22 +52,7 @@ heapless = "0.7"
...
-    "python_array_test",
-    "slugs",
-    "standard",
-    "test",
-    "ualberta",
-    "uavionix",
-    "icarous",
-    "common",
-]
+"all-dialects" = ["ardupilotmega", "asluav", "autoquad", "matrixpilot", "minimal", "paparazzi", "python_array_test", "slugs", "standard", "test", "ualberta", "uavionix", "icarous", "common"]
```
patrickelectric: do not follow up style changes with funcional changes as well, having such list in a huge single line is harder to maintain.

patrickelectric (changes requested):
Just some unnecessary style change to approve this, everything else looks awesome.

danieleades:
> I would like to see how this does after #161 gets merged. Maybe the 1.84s of rayon comp time will already make it not worth it.

I agree it's worth merging your changes first and then reassessing.

I think also it maybe be worth considering initial compilation and subsequent compilation separately. For example maybe its worth compiling rayon *once* to get a benefit on every subsequent recompilation of this library

Cargo.toml
danieleades: no problem. This was introduced as an artefact of using `cargo edit` to add the rayon dependency (which can change formatting), not an intentional change

GabrielDertoni:
Another way to significantly improve developer experience for this crate is using a separate crate for codegen, and letting it run `build/parse.rs` and including the generated code. This way, developers that don't want to change the parser, don't need to wait for the build script to rerun and all the exact same generated code get compiled once again.

PR #164, opened 2023-03-30
Update mavlink submodule

### LINKED ISSUE #141: Update Mavlink submodule
The mavlink submodule should be updated, there is a few things in the newer versions which does not work with the parsing.
I started on it in https://github.com/mavlink/rust-mavlink/pull/132, but never finished it.

lopsided98:
This PR enables code generation for bitmask enum arrays, which allows the mavlink submodule to be updated. Bitmask enum array fields are now handled in the same way as non-bitmask enum arrays; they are just generated as primitive arrays rather than actual enums. This is not ideal, but fixing it for real probably requires either some special case code for enum arrays or adjusting the parser architecture.

Note that there currently only appears to be one case of a bitmask enum array (`ESC_INFO.failure_flags`) and one case of a non-bitmask enum array (`TRAJECTORY_REPRESENTATION_WAYPOINTS.command`).

Fixes #141 

cc @Erk-

### DIFF
--- a/build/parser.rs
+++ b/build/parser.rs
@@ -66,7 +66,7 @@ impl MavProfile {
                             for mut enm in self.enums.values_mut() {
                                 if enm.name == *enum_name {
                                     // this is the right enum
-                                    enm.bitfield = Some(field.mavtype.rust_type());
+                                    enm.bitfield = Some(field.mavtype.rust_primitive_type());
                                 }
                             }
                         }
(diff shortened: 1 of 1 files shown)


patrickelectric:
That, great, I'll check it out asap.

Erk-:
I think you also need to add the new dialects to the `Cargo.toml` file and the workflow. As far as I can tell the new ones are `storm32` and `development`.

patrickelectric:
I would say that's fine if the others protocol do not work, this is already a major win.

patrickelectric:
@lopsided98 can you fix the style for us to merge ?

patrickelectric (approved):

patrickelectric:
I'm just going to test the API with serde to be sure that nothing broke before merging if that is ok

PR #170, opened 2023-04-10
Add Raspi Example

BWStearns:
Added a raspi -> pixhawk4 example.

### DIFF
--- a/examples/raspi/Cargo.toml
+++ b/examples/raspi/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "raspi_mavlink_example"
+version = "0.1.0"
+edition = "2021"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+rppal = "0.14.1"
+mavlink = "0.10.1"
--- a/examples/raspi/src/main.rs
+++ b/examples/raspi/src/main.rs
@@ -1,0 +1,122 @@
+use std::error::Error;
+use std::f32::consts::PI;
+use std::{env, sync::Arc, thread, time::Duration};
+
+use rppal::pwm::{Channel, Polarity, Pwm};
+use mavlink::error::MessageReadError;
+use mavlink::common::MavMessage;
+use mavlink;
+
+// Servo configuration. The datasheet was wrong for mine so these are manually calibrated.
+// Your configuration may be different.
+const PERIOD_MS: u64 = 20;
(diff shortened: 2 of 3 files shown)


BWStearns:
The example reads out Mavlink on Serial and prints all the output as well as moving a servo in response to attitude data. I think it's helpful in terms of showing a bunch of the message formats and a bit of interacting with the data received.

examples/raspi/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+use mavlink::error::MessageReadError;
+use mavlink::common::MavMessage;
+use mavlink;
+
+// Servo configuration. The datasheet was wrong for mine so these are manually calibrated.
+// Your configuration may be different.
+const PERIOD_MS: u64 = 20;
+const PULSE_MIN_US: u64 = 640;
+const PULSE_NEUTRAL_US: u64 = 1540;
+const PULSE_MAX_US: u64 = 2540;
```
onur-ozkan, line 15: I think we should provide a more generic and compatible example rather than telling people to adjust these values to fit their system.

BWStearns: We could just log stuff out I guess?

onur-ozkan: That makes sense. It's also worth mentioning which RP model this was tested on and confirmed to work.

examples/raspi/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "raspi_mavlink_example"
```
onur-ozkan, line 2: To match the dir name:

```suggestion
name = "raspi"
```

onur-ozkan (commented):
It's worth to rebase the PR with latest master as it's based on extremely old commit right now.

Can't get a working build on my local:

```sh
$ remotes/BWStearns/raspi-example $ cargo build
warning: `raspi/.cargo/config` is deprecated in favor of `config.toml`
note: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
    Updating crates.io index
error: failed to select a version for the requirement `xml-rs = "^0.2"`
  version 0.2.0 is yanked
  version 0.2.1 is yanked
  version 0.2.2 is yanked
  version 0.2.3 is yanked
  version 0.2.4 is yanked
location searched: crates.io index
required by package `mavlink v0.10.1`
    ... which satisfies dependency `mavlink = "^0.10.1"` of package `raspi v0.1.0 (rust-mavlink/examples/raspi)`
```

PR #193, opened 2023-09-10
String parsing for char[]

gbin:
This allows messages like STATUSTEXT to parse nicely as an ArrayString in Rust.


patrickelectric (changes requested):
The PR appears to break the embedded support. 
The second commit is also fixing the first one, so they should be merged.

patrickelectric:
If having embedded support is not possible, we may turn this as a feature to be used in compite time.

onur-ozkan:
Could you resolve conflict please?

PR #201, opened 2023-10-22
refactor 'internal tests' CI job to use github matrix

danieleades:
use a github action 'matrix' rather than hand-rolling with a shell script

also removed the 'autoquad' feature, which doesn't seem to exist (the old script hid the error)


patrickelectric (approved):
Thank you very much!

PR #204, opened 2023-11-20
Fix sequence order in serialization function

IhsenBouallegue:
I am currently doing my bachelor's degree about MAVLink and LoRa. I decided to use rust and therefore rust-mavlink.
I need to serialize and deserialize MavFrames to send them over the LoRa link. While doing that, I noticed a weird behavior of the deser function. The order of serializing the header fields (system_id, component_id and sequence) does not align with the order of deserialization. 

Here is a screenshot showing the problem
image

Order of serialization
image

Old order of deserialization
image

Also, if you look closely, the values are all over the place and do not match (it's not a corrupt message problem). I am going to open a separate issue for that (https://github.com/mavlink/rust-mavlink/issues/205), as I am still investigating it and is not related to the header.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,9 +154,9 @@ impl<M: Message> MavFrame<M> {
         let mut buf = bytes_mut::BytesMut::new(buf);
 
         // serialize header
+        buf.put_u8(self.header.sequence);
         buf.put_u8(self.header.system_id);
         buf.put_u8(self.header.component_id);
-        buf.put_u8(self.header.sequence);
 
         // message id
         match self.protocol_version {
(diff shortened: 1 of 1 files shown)


patrickelectric (changes requested):
Hi @IhsenBouallegue, thanks for identifying the problem! 
But the issue is in the serialization and not on deserialization.
The order behind it is explained here:
https://mavlink.io/en/guide/serialization.html

IhsenBouallegue:
> Hi @IhsenBouallegue, thanks for identifying the problem! But the issue is in the serialization and not on deserialization. The order behind it is explained here: https://mavlink.io/en/guide/serialization.html

Oh I blindly assumed that the serialization is the right one. I will switch that one then. Thanks for the catch!

patrickelectric:
Fix #205

PR #208, opened 2023-11-21
0.10: Fix serialization and deserialization problems

patrickelectric:
Based on #204 #206 #207

### DIFF
--- a/tests/mav_frame_tests.rs
+++ b/tests/mav_frame_tests.rs
@@ -23,7 +23,7 @@ mod mav_frame_tests {
     ];
 
     #[test]
-    pub fn test_deser() {
+    pub fn test_deser_ser() {
         use mavlink::{common::MavMessage, MavFrame, MavlinkVersion};
         let frame = MavFrame::<MavMessage>::deser(MavlinkVersion::V2, HEARTBEAT_V2)
             .expect("failed to parse message");
@@ -31,9 +31,11 @@ mod mav_frame_tests {
         assert_eq!(frame.header, crate::test_shared::COMMON_MSG_HEADER);
         let heartbeat_msg = crate::test_shared::get_heartbeat_msg();
 
-        let msg = match frame.msg {
-            MavMessage::HEARTBEAT(msg) => msg,
-            _ => panic!("Decoded wrong message type"),
+        let buffer = frame.ser();
+        assert_eq!(buffer, HEARTBEAT_V2[..buffer.len()]);
+
+        let MavMessage::HEARTBEAT(msg) = frame.msg else {
+            panic!("Decoded wrong message type");
         };
         assert_eq!(msg.custom_mode, heartbeat_msg.custom_mode);
         assert_eq!(msg.mavtype, heartbeat_msg.mavtype);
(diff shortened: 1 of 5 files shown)


tests/mav_frame_tests.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub fn test_deser_ser() {
+        use mavlink::{common::MavMessage, MavFrame, MavlinkVersion};
+        let frame = MavFrame::<MavMessage>::deser(MavlinkVersion::V2, HEARTBEAT_V2)
+            .expect("failed to parse message");
+
+        assert_eq!(frame.header, crate::test_shared::COMMON_MSG_HEADER);
+        let heartbeat_msg = crate::test_shared::get_heartbeat_msg();
+
+        let buffer = frame.ser();
+        assert_eq!(buffer[..buffer.len() - 2], HEARTBEAT_V2[..buffer.len() - 2]);
```
joaoantoniocardoso: we are.. skipping the checksum? reasons?

patrickelectric: > we are.. skipping the checksum? reasons?

we do not calculate the checksum on the serialization function

joaoantoniocardoso: Ok, but then shouldn't it be `assert_eq!(buffer, HEARTBEAT_V2[..buffer.len()]);` instead? Otherwise, we'll be skipping more than that. Also, it'd be good to add a note that this is the reason, it's not that obvious.

patrickelectric: Yep, thanks for finding it!

tests/mav_frame_tests.rs
```diff
@@ -0,0 +1,97 @@
...
+        assert_eq!(frame.header, crate::test_shared::COMMON_MSG_HEADER);
+        let heartbeat_msg = crate::test_shared::get_heartbeat_msg();
+
+        let buffer = frame.ser();
+        assert_eq!(buffer, HEARTBEAT_V2[..buffer.len()]);
+
+        let msg = match frame.msg {
+            MavMessage::HEARTBEAT(msg) => msg,
+            _ => panic!("Decoded wrong message type"),
+        };
```
joaoantoniocardoso: not important, but if the minimum Rust version allows, it could be written as:
```Rust
        let MavMessage::HEARTBEAT(msg) = frame.msg else {
            panic!("Decoded wrong message type");
        };
```

tests/mav_frame_tests.rs
```diff
@@ -0,0 +1,97 @@
...
+        assert_eq!(msg.custom_mode, heartbeat_msg.custom_mode);
+        assert_eq!(msg.mavtype, heartbeat_msg.mavtype);
+        assert_eq!(msg.autopilot, heartbeat_msg.autopilot);
+        assert_eq!(msg.base_mode, heartbeat_msg.base_mode);
+        assert_eq!(msg.system_status, heartbeat_msg.system_status);
+        assert_eq!(msg.mavlink_version, heartbeat_msg.mavlink_version);
+    }
+
+    #[test]
+    pub fn test_deser_ser_message() {
```
joaoantoniocardoso, line 53: This test would be awesome with [proptest](https://github.com/proptest-rs/proptest)

joaoantoniocardoso (approved):
Seems okay!

Issue #246, opened 2024-07-30
Mavlink v2 extension are broken for v0.13 for the mavlink crate


pv42:
The feature `emit-extensions` that is supposed to enable mavlink v2 extension does not currently work for versions 0.13.x. 
This can be seen when comaparing the documentation (which enables all features) of version 0.12.2 and 0.13.1 on docs.rs for COMMAND_ACK_DATA: 
https://docs.rs/mavlink/0.13.1/mavlink/common/struct.COMMAND_ACK_DATA.html only has the `command` and `result` fields
while https://docs.rs/mavlink/0.12.2/mavlink/common/struct.COMMAND_ACK_DATA.html also contains the extension fields.

The reason for this is that the feature when enabled on the `mavlink` crate will not be set on the `mavlink-bindgen`crate which uses it when generating code.
This can be fixed by adding the feature dependency in `mavlink/Cargo.toml`:
```diff
@@ -88,7 +88,7 @@ serde_arrays = { version = "0.1.0", optional = true }
 
 "format-generated-code" = []
 "emit-description" = []
-"emit-extensions" = []
+"emit-extensions" = ["mavlink-bindgen/emit-extensions"]
 "std" = ["mavlink-core/std"]
 "udp" = ["mavlink-core/udp"]
 "tcp" = ["mavlink-core/tcp"]
```
and adding the feature itself to mavlink-bindgen in `mavlink-bindgen/Cargo.toml`:
```diff
@@ -27,3 +27,4 @@ anstyle-parse = { version = "=0.2.1", optional=true }
 
 [features]
 cli = ["dep:clap", "dep:clap_lex", "dep:clap_builder", "dep:anstyle", "dep:anstyle-query", "dep:anstyle-parse"]
+emit-extensions=[]
```
There is a further issue when doing that, since the parser will use `serde(default = "crate::RustDefault::rust_default"))` and RustDefault is a `pub (crate)` is a trait in `mavlink-core` but required in the `mavlink` crate when using `emit-extensions`.
The solution to this is making it fully public and importing it in `mavlink/lib.rs`. This does not create additional dependencies since mavlink already depends on mavlink-bindgen.

```diff
@@ -3,3 +3,6 @@
 include!(concat!(env!("OUT_DIR"), "/mod.rs"));
 
 pub use mavlink_core::*;
+
+#[cfg(feature = "emit-extensions")]
+#[allow(unused_imports)]
+pub (crate) use mavlink_core::utils::RustDefault;
```

[...]

Issue #165, opened 2023-04-01
Add new dialects from Mavlink upstream


Erk-:
The upstream version was updated in #164, but it did not add new dialects to the feature set.

The new dialets as of time writing this are `storm32`, `AVSSUAS`, `all`, `cubepilot` and `development`.

And the dialect `autoquad` was removed.

patrickelectric:
The only one missing now is storm32 that is having problems with the bitflag enum

patrickelectric:
https://github.com/mavlink/mavlink/issues/1969

PR #248, opened 2024-08-03
Emit extensions

pv42:
Fixes emit-extensions and emit-descriptions features not being used by mavlink-bindgen (#246).
This requires a (none breaking) bump in version since fixes in the mavlink crate require newer versions of mavlink-bindgen/mavlink-core crates.
To ensure test work this required adding the path specifications for the dependencies within this crate.

Fixes missing storm32 dialect feature.

### DIFF
--- a/mavlink/Cargo.toml
+++ b/mavlink/Cargo.toml
@@ -64,7 +64,6 @@ serde_arrays = { version = "0.1.0", optional = true }
 "ualberta" = ["common"]
 "uavionix" = ["common"]
 "icarous" = []
-"storm32" = []
 "common" = []
 "cubepilot" = ["common"]
 
@@ -85,7 +84,6 @@ serde_arrays = { version = "0.1.0", optional = true }
     "icarous",
     "common",
     "cubepilot",
-    "storm32",
 ]
 
 "format-generated-code" = []
--- a/mavlink/Cargo.toml
+++ b/mavlink/Cargo.toml
@@ -49,6 +49,7 @@ serde_arrays = { version = "0.1.0", optional = true }
     "uavionix",
     "avssuas",
     "cubepilot",
+    "storm32",
 ]
 "ardupilotmega" = ["common", "icarous", "uavionix"]
 "asluav" = ["common"]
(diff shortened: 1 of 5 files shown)


mavlink/Cargo.toml
```diff
@@ -19,15 +19,15 @@ edition = "2018"
...
-mavlink-bindgen = { version = "0.13.1", default-features = false }
+mavlink-bindgen = { path = "../mavlink-bindgen", version = "0.13.2", default-features = false }
 
 [[example]]
 name = "mavlink-dump"
 path = "examples/mavlink-dump/src/main.rs"
 required-features = ["ardupilotmega"]
 
 [dependencies]
-mavlink-core = { version = "0.13.1", default-features = false }
```
patrickelectric: You should avoid the following for the release system to work.

patrickelectric (changes requested):

mavlink/Cargo.toml
pv42: I have removed the versions in 33b4d4ec3c971bb1af81080c31a77304a790096e but this breaks most tests.
Looking at the failed build log of 9cd7bf0b8c720d255232a389051fb60f5fc92c2c before the path dependencies were removed the issue seems to be that no version at all is specified:
> error: all dependencies must have a version specified when publishing.

But since cargo supports specifying a [path and a version](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) doing so should circumvent the issue.

amsmith-pro:
Looking forward to this as well, thank you

pv42:
I was originally under the impression storm32 was just overlocked and since it generated a compiler warning I just added it.
But #165 indicated it was intentionally left out. I also probably should never have been in this unrelated PR. Therefore I removed it.

mavlink/Cargo.toml
pv42: resolved based with merged PR #255 that revert to using paths

patrickelectric (approved):
Thanks! After brainstorming more we decided to continue with the path approach and fix the release method in a different way.

patrickelectric:
@pv42 can you rebase over master ? It appears to contain conflicts.

pv42:
Fixed the conflicts with 0fb0e2e.

pv42:
Should have be properly rebased now.

Issue #82, opened 2021-01-27
Don't support signing?


hamishwillee:
Just confirming for the mavlink docs tat this library doesn't support signing - right?

Are there any more docs than provided on the readme? 
For context, I'm trying to characterize the things provided by each project
- ie level of docs (where [this one for C](https://mavlink.io/en/mavgen_c/) might be full
- level of testing - ie what is tested, unit test, round trip tests, test that send receive from C library etc.

patrickelectric:
Hi @hamishwillee, I hope that you are doing fine.

About rust-mavlink, it indeed has a support for signed messages, but not fully.
It can receive but will do nothing with the signed value itself.

For documentation, you can get it here: https://docs.rs/mavlink/latest/mavlink/
For testing: https://github.com/mavlink/rust-mavlink/tree/master/tests
    - They are done with some hardcoded messages to check if the pack/unpack values are correct and a self test of the internal functions.

hamishwillee:
Thank you very much for the clarification.

hamishwillee:
Note, I've updated mavlink docs with link in https://github.com/mavlink/mavlink-devguide/pull/312

I have not updated the information on signing because it still isn't quite clear to me what you mean by not supported. Full signing support would look something like:
1. Has a mechanism for setting a key to use for signing messages (key management itself might be in hardware).
1. Has mechanism for turning signing on and off
1. If signing turned on will sign MAVLink 2 messages.
1. If signing not turned on will parse signed and unsigned messages
1. If signing turned on will reject unsigned messages by default, but has hooks to allow an implementation to choose to keep some messages.

Assuming we're aligned on the above can you confirm again which it does/does not do?

tjrmarques:
Hi!
It's been a while since this has been merged, any idea on when it will land in crates.io?

patrickelectric:
@tjrmarques it appears that the release process has changed on rust-mavlink and more work on that is now necessary to do the release, I'm out of time at the moment.

pv42:
@hamishwillee the newest release now supports signing to the following extend:

1. Has a mechanism for setting a key to use for signing messages (key management itself might be in hardware).
Yes

2. Has mechanism for turning signing on and off
Yes

3. If signing turned on will sign MAVLink 2 messages.
Yes (can be disabled to only receive signed messages send unsigned ones)

4. If signing not turned on will parse signed and unsigned messages
Yes

5. a) If signing turned on will reject unsigned messages by default,
 Yes
[...]

PR #253, opened 2024-08-11
Signing

pv42:
As pointed out in #82 [MAVLink 2 message signing](https://mavlink.io/en/guide/message_signing.html) is not supported beyond being able to correctly parse signed messages.
This PR adds full support for signature creation and verification. 
This is done by introducing the `signing` feature to the `mavlink` and `mavlink-core` crates.
The implementation is done in a way that the public interfaces remain unchanged without the feature and are semver compatible when it is enabled.

# Additions
- `signing` feature for mavlink-core and as passthrough to `mavlink`, all following additions require this, this is not enabled by default 
- add `SigningConfig` to configure signing with a secret key, enable sign outgoing and allow unsigned messages
- add `SigningData` to execute signing with `verify_signature()`, `sign_message()`
- `MavConnection::setup_signing(..)` to all connection implementations
- add signing versions of `pub fn` that have an additional `Option<&SigningData>` paramter: `read_versioned_msg_signed()`, `read_v2_raw_message_signed()`, `read_v2_msg_signed()`, `write_versioned_msg_signed()`, `write_v2_msg_signed()`
- add byte access function to `MAVLinkV2MessageRaw`: `checksum_bytes()`, `signature_link_id[_mut]()`, `signature_timestamp[_bytes[_mut]]()`, `signature_value[_mut]()`
- fix `MAVLinkV2MessageRaw::header()` requiring a mut reference
- add `MAVLinkV2MessageRaw::calculate_signature()` that calculates a messages current signature
- add `MAVLinkV2MessageRaw::serialize_message_for_signing()` to serialze while setting the incompat_flag

# Signing specifications
With the `signing` feature enabled and signing setup the following applies:
- All incoming messages with invalid signature are ignored
- All incoming messages that are not signed are ignored unless `allow_unsigned` is set
- All outgoing message are signed when `sign_outgoing` is set.
- If the message signature does not match the message is ignored.
- If a messages timestamp is more then 60s older then the newest received message it is ignored
- If a messages timestamp is older then the previous message from the same (system,component,link) it is ignored.
- Timestamps are strictly monotonic increasing per send message allowing it get ahead of real time.

# Tests
- add signing test to verify the signing itself
- add signing to tcp_loopback_tests to check the connection signing implementation
- enable `signing` feature for all test

# Caveats
- The mavlink standart calls for a method to allow individuall unsigned messages, this currently has only a simple boolean flags. A better solution is to either allow messages ids based on a whitelist or a callback.
- The link id used is fixed to 0. The standart describes that this should be different for each connection from this device, however this is difficult to implement and I could not come up with a satisfying solution.
- The timestamps required for signing are generated using `std::SystemTime::now()` which might not exist in a none-std enviroment, maybe the `signing` feature should depend on `std`.
- This PR suffers from the same issue as #246 as test on `mavlink` are broken since they depend on the published version of `mavlink-core` which does not have signing, when testing locally with a path dependency they succeed.
- The signing feature uses the `incompat_flags` field of the MAVLink header, for the purpose of preserving API stability the coresponding struct `MavHeader` is not modified but this should probably be done in a future breaking version.
- It is possible to enable signing on a `MavConnection` and then set a it to use MAVLink 1 only which ignores the setup signing configuration.


patrickelectric:
Can you rebase over master ?

pv42:
I have fixed the merge conflicts for this, but I should probably add signing variant of the new async methods too. Hence I marked this a draft until I have implemented this.

pv42:
I have added `read_v2_raw_message_async_signed`, added `link_id` to signing setup and slightly modified tests.

joaoantoniocardoso:
Nice work, but I think you've merged the updated master into this branch instead of rebasing it over master

.github/workflows/test.yml
```diff
@@ -59,7 +59,7 @@ jobs:
         with:
           use-cross: true
           command: check
-          args: --all --all-targets
+          args: --all --all-targets --features signing
```
patrickelectric: Please add a test without signing to be sure that it stills compile

mavlink-core/src/connection/udp.rs
```diff
@@ -143,7 +160,24 @@ impl<M: Message> MavConnection<M> for UdpConnection {
...
+                #[cfg(feature = "signing")]
+                self.signing_data.as_ref(),
+            )?;
+            #[cfg(feature = "signing")]
+            write_versioned_msg_signed(
+                &mut buf,
+                self.protocol_version,
+                header,
+                data,
+                #[cfg(feature = "signing")]
```
patrickelectric: Does this make sense ?

.github/workflows/test.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - uses: actions/checkout@master
       - uses: dtolnay/rust-toolchain@stable
       - name: Run internal tests
-        run: cargo test --verbose --features ${{ matrix.dialect }} -- --nocapture
+        run: cargo test --verbose --features ${{ matrix.dialect }} --features signing -- --nocapture
```
patrickelectric: Please add it without signing as well

patrickelectric (changes requested):

mavlink-core/src/connection/udp.rs
pv42: No it did not.

patrickelectric (approved):

PR #261, opened 2024-08-25
Minor refactoring

danieleades:
just some tiny nits.

This should *possibly* be merged after #262, as there may be a couple of further tweaks possible with the edition change


joaoantoniocardoso (approved):
nice!

patrickelectric (approved):

PR #267, opened 2024-08-25
chore(deps): update 'embedded_hal'

danieleades:
does CI actually cover this feature?


patrickelectric:
> does CI actually cover this feature?

Yes https://github.com/mavlink/rust-mavlink/actions/runs/10548211116/job/29221931254

danieleades:
ah i think i see what's going on here. the `embedded_hal_02` dependency is essentially a feature which pins to an older version. Therefore this PR should not be merged.

when `embedded-io` provides feature parity the `embedded_hal_02` dependency/feature should be removed, not updated.

Issue #256, opened 2024-08-22
Async Connection Trait


pv42:
With #255 support for `tokio`-based `async` was merged, adding `async` versions of `write_versioned_msg` and `read_v2_msg` which is very welcome (as it is something I was thinking of doing myself).

I suggest adding an `async` version of the connection trait similar to `MavConnection<M: Message>`: 
```rs
#[async_trait]
pub trait AsyncMavConnection<M: Message> {
    async fn recv(..) -> ..
    async fn send(..) -> ..
    fn set_protocol_version(..)
    fn get_protocol_version(..) -> ..
    async fn send_frame(..) -> ..
    async fn recv_frame(..) -> .. 
    async fn send_default(..) -> ..
}
```
And then implementing it for the `tcp`, `udp`, `file` and potentially `serial` (unsure if/how `tokio` support that) connections and providing a public function to open `async` connections:
`pub async fn connect_async<M: Message>(address: &str) -> io::Result<Box<dyn AsyncMavConnection<M> + Sync + Send>>`

If this is something that would fit and if @joaoantoniocardoso is not already working on it I would be happy to submit a PR.

joaoantoniocardoso:
Hi @pv42, nice idea, surely [mavlink-camera-manager](https://github.com/mavlink/mavlink-camera-manager) (and potentially [mavlink2rest](https://github.com/mavlink/mavlink2rest)) would benefit from that!

To contextualize, I did implement these tokio-based async read/write because I've been experimenting with a new mavlink router written in rust ([mavlink-server](https://github.com/bluerobotics/mavlink-server)). This router can also be refactored to make use of that, but it already has an implementation (which might help us here?).

Feel free to use any of that code if useful, the mavlink-server repo still lacks a license file, but it's intended to be MIT.

BWStearns:
I was literally looking to see if this existed last night! +100!

BWStearns:
Given #255 what does an async connection trait look like broad strokes? Like what would a ticket look like for this?

joaoantoniocardoso:
@pv42 just so you know, I'm reworking the tokio async reader around tokio-utils' codec, so we can ensure a frame would be received from the raw read functions, and using the same codec you can have a tokio's UdpStream.

PR #276, opened 2024-08-29
async connections

pv42:
Implements #256.
Adds asynchronous connect method `async connect_async(address)` as async version of `connect(address)`.
For this the following is added when the `tokio-1` feature is enabled:
- `mod async_connection` with `connect_async(..)` 
- `trait AsyncMavConnection` as `[#[async_trait]` as an async verions of `trait MavConnection`
- `AsyncFileConnection`, `AsyncTcpConnection`, `AsyncUdpConnection` implementing `AsyncMavConnection`
- `AsyncPeekReader` as async verions of `PeekReader`
- read/write functions `read_versioned_msg_async`, `read_versioned_msg_async_signed`, `read_v1_raw_message_async`, `read_v1_msg_async`, `read_v2_msg_async_signed`, `write_versioned_msg_async_signed`, `write_v2_msg_async_signed`
- change functions `read_v2_raw_message_async`, `read_v2_raw_message_async_inner`, `read_v2_raw_message_async_signed`, `read_v2_msg_async` to read from a `AsyncPeekReader` instead of directly for a `tokio::io::AsyncReadExt`
- `connect_async` is exposed as `mavlink::connect_async`

Tests:
- Add async versions of tcp_loopback_tests.rs and udp_loopback_async_tests.rs
- `AsyncUdpConnection` tests datagram_buffering similar to its blocking sibling
- To run test the `tokio` features `rt`, `macros` are needed, for this reason `tokio` with these additional features is required as `dev-dependencies` for both `mavlink-core` and `mavlink`

Caveats 
- No serial support.
- The change of some existing function to `AsyncPeekReader` from `tokio::io::AsyncReadExt` is breaking
- The MSRV test fails (see #277/#278), test succeds locally for bins/lib targets or without using no-dev-deps.
- The combinations of asnyc and signing results in 4 versions of the same functions that effectifly do the same. (e.g. `write_versioned_msg`, `write_versioned_msg_signed`, `write_versioned_msg_async` and `write_versioned_msg_async_signed`)

### DIFF
--- a/mavlink-core/Cargo.toml
+++ b/mavlink-core/Cargo.toml
@@ -26,7 +26,7 @@ embedded-io-async = { version = "0.6.1", optional = true }
 serde = { version = "1.0.115", optional = true, features = ["derive"] }
 serde_arrays = { version = "0.1.0", optional = true }
 serial = { version = "0.4", optional = true }
-tokio = { version = "1.0", default-features = false, features = ["io-util", "net", "sync", "fs", ], optional = true }
+tokio = { version = "1.0", default-features = false, features = ["io-util", "net", "sync", "fs"], optional = true }
 sha2 = { version = "0.10", optional = true }
 async-trait = { version = "0.1.18", optional = true }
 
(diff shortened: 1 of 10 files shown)


patrickelectric (approved):
Nice!

Issue #277, opened 2024-09-04
Test with no-dev-deps all-targets


pv42:
The MSRV check tests run
`cargo no-dev-deps check --all --all-targets [features]`
this appears to me as a mistake since no-dev-deps disables dev dependencies but --all-targets includes test which (might) need dev dependencies.
This was changed recently by @danieleades in a729400 so I was wondering if there is a reason for it that I am missing or if is just a mistake. If so it can by fixed by changing the check to either `cargo check --all --all-targets [features]` or split commands for targets that need dev-deps:
- `cargo no-dev-deps check --all --lib --bins [features]`
- `cargo check --all --tests --benches --examples [features]`

danieleades:
Yeah I suspect `--all-targets` can be removed

PR #278, opened 2024-09-04
Run MSRV check only for lib and bin targtets.

pv42:
See #277.
Removes tests, examples and benches (there are currently no benches) from msrv test.
Since the msrv test runs without dev dependencies they would potentially not work.


patrickelectric:
@pv42 but it's already passing, what is the point of removing it ?

pv42:
It is a reason why the CI on #276 fail. 
It would also cause potential issues if anyone would like to add any dev dependencies at some point.

patrickelectric (approved):

PR #280, opened 2024-09-10
fix dependabot config

danieleades:
looks like i introduced an error in the dependabot config. This PR fixes that error


joaoantoniocardoso (approved):

onur-ozkan:
What is the error?

danieleades:
> What is the error?

The config file is structured incorrectly

PR #284, opened 2024-09-21
feat: async serial

roby2014:
Tried following the other connection wrappers structures.
Needs testing (i'll test soon).

### DIFF
--- a/mavlink-core/src/async_connection/mod.rs
+++ b/mavlink-core/src/async_connection/mod.rs
@@ -8,6 +8,9 @@ mod tcp;
 #[cfg(feature = "udp")]
 mod udp;
 
+#[cfg(feature = "direct-serial")]
+mod direct_serial;
+
 mod file;
 
 #[cfg(feature = "signing")]
@@ -70,9 +73,9 @@ pub trait AsyncMavConnection<M: Message + Sync + Send> {
 ///  * `udpin:<addr>:<port>` to create a UDP server, listening for incoming packets
 ///  * `udpout:<addr>:<port>` to create a UDP client
 ///  * `udpbcast:<addr>:<port>` to create a UDP broadcast
+///  * `serial:<port>:<baudrate>` to create a serial connection
 ///  * `file:<path>` to extract file data
 ///
-/// Serial is currently not supported for async connections, use [`crate::connect`] instead.
 /// The type of the connection is determined at runtime based on the address type, so the
 /// connection is returned as a trait object.
 pub async fn connect_async<M: Message + Sync + Send>(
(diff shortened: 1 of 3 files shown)


roby2014:
Did a simple test with `serial:/dev/ttyACM1:115200`:
```rs
// ...
        let mut connection =
            mavlink::connect_async::<mavlink::ardupilotmega::MavMessage>("serial:/dev/ttyACM1:115200").await?;
        connection.set_protocol_version(self.mavlink_version);
// ...
            select! {
                res = connection.recv() => {
                    match res {
                        Ok((header, msg)) => {
                            trace!(?header, ?msg, "received message from connection");
                            //...
                        }
                        Err(e) => {
                            error!("failed to read: {e}");
                        }
                    }
                }
                res = channel.recv() => {
                    // ...
                }
            }
```
and I'm able to get data:
```
received message from connection header=MavHeader { system_id: 1, component_id: 1, sequence: 133 } msg=ATTITUDE(ATTITUDE_DATA { time_boot_ms: 389342, roll: 0.0, pitch: 0.0, yaw: 0.0, rollspeed: 0.0, pitchspeed: 0.0, yawspeed: 0.0 })

received message from connection header=MavHeader { system_id: 1, component_id: 1, sequence: 134 } msg=VFR_HUD(VFR_HUD_DATA { airspeed: 0.0, groundspeed: 0.0, alt: 0.0, climb: 0.0, heading: 0, throttle: 0 })
```

mavlink-core/src/async_connection/mod.rs
```diff
@@ -70,6 +73,7 @@ pub trait AsyncMavConnection<M: Message + Sync + Send> {
 ///  * `udpin:<addr>:<port>` to create a UDP server, listening for incoming packets
 ///  * `udpout:<addr>:<port>` to create a UDP client
 ///  * `udpbcast:<addr>:<port>` to create a UDP broadcast
+///  * `serial:<port>:<baudrate>` to create a serial connection
 ///  * `file:<path>` to extract file data
 ///
 /// Serial is currently not supported for async connections, use [`crate::connect`] instead.
```
pv42: You should probably remove the line that says serial is not supported

roby2014: Done! :)

peterkrull:
I would also like to have async serial be upstreamed so there is consistency between sync/async. Though looking at your code, is it not problem that the port is locked behind a single mutex, i.e. no reader and writer half? If I were to wrap the AsyncSerialConnection in an Arc and create a reader and writer thread, would the reader not be able to lock out the writer indefinitely if there are no incoming messages?

patrickelectric (approved):

PR #286, opened 2024-09-24
implement `recv_raw` for Connection wrappers into `mavlink-core`

roby2014:
(depends on #284)

So, my goal is to have some sort of way to fetch raw messages from the connection wrappers ([`AsyncMavConnection`](https://github.com/mavlink/rust-mavlink/blob/master/mavlink-core/src/async_connection/mod.rs), [`MavConnection`](https://github.com/mavlink/rust-mavlink/blob/master/mavlink-core/src/connection/mod.rs#L20)) directly, without having to do the extra-step of providing a type and converting between and vice-versa.

At the moment, `recv` will always use the generic `M`.

Another thing I was planning was to somehow allow connections without knowing the message type `M`, but this looks a bit harder to implement (and probably impossible because dialect must be known at compile-time?).

I've already implemented for `AsyncMavConnection` and its protocols (`udp`, `tcp`, `direct-serial`, `file`):
```rs
async fn recv_raw(&self) -> Result<MAVLinkRawMessage, crate::error::MessageReadError>;
``` 
I've also implemented a small example (`mavlink-dump-raw-async`), which already shows how this can be used with `async`.

**I want to ask if this is a good idea and if possible, to open discussion on this topic here.**

### DIFF
--- a/mavlink-core/Cargo.toml
+++ b/mavlink-core/Cargo.toml
@@ -29,6 +29,7 @@ serial = { version = "0.4", optional = true }
 tokio = { version = "1.0", default-features = false, features = ["io-util", "net", "sync", "fs"], optional = true }
 sha2 = { version = "0.10", optional = true }
 async-trait = { version = "0.1.18", optional = true }
+tokio-serial = { version = "5.4.4", default-features = false, optional = true }
 
 [features]
 "std" = ["byteorder/std"]
@@ -41,7 +42,7 @@ async-trait = { version = "0.1.18", optional = true }
 "embedded" = ["dep:embedded-io", "dep:embedded-io-async"]
 "embedded-hal-02" = ["dep:nb", "dep:embedded-hal-02"]
 "serde" = ["dep:serde", "dep:serde_arrays"]
-"tokio-1" = ["dep:tokio", "dep:async-trait"]
+"tokio-1" = ["dep:tokio", "dep:async-trait", "dep:tokio-serial"]
 "signing" = ["dep:sha2"]
 default = ["std", "tcp", "udp", "direct-serial", "serde"]
 
--- a/mavlink-core/Cargo.toml
+++ b/mavlink-core/Cargo.toml
@@ -26,7 +26,12 @@ embedded-io-async = { version = "0.6.1", optional = true }
 serde = { version = "1.0.115", optional = true, features = ["derive"] }
 serde_arrays = { version = "0.1.0", optional = true }
 serial = { version = "0.4", optional = true }
-tokio = { version = "1.0", default-features = false, features = ["io-util", "net", "sync", "fs"], optional = true }
+tokio = { version = "1.0", default-features = false, features = [
+    "io-util",
+    "net",
+    "sync",
+    "fs",
+], optional = true }
 sha2 = { version = "0.10", optional = true }
 async-trait = { version = "0.1.18", optional = true }
 tokio-serial = { version = "5.4.4", default-features = false, optional = true }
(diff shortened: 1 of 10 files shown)


roby2014:
Well 1 possible idea I have is to implement `RawConnectors`, e.g: `AsyncMavRawConnection` and `MavRawConnection` instead of adding `recv_raw`. This would not care about dialect and just tries to parse a MAVLink message. @patrickelectric (and other maintainers) what do you think?

pv42:
So I had the idea of doing something like that:
```rust
pub trait MavRawConnection {
    fn recv_raw(&self) -> Result<MAVLinkVXMessageRaw>
    ...
}
```
and then have 
```rust
pub trait MavConnection<M: Message> : MavRawConnection  {
    fn recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError> {
        M::parse(self.recv_raw())
    }
   ...
}
```
The issue is that `MAVLinkVXMessageRaw` does not exist and it probably would have to be a trait object.

roby2014:
> So I had the idea of doing something like that:
> 
> ```rust
> pub trait MavRawConnection {
>     fn recv_raw(&self) -> Result<MAVLinkVXMessageRaw>
>     ...
> }
> ```
> 
> and then have
> 
> ```rust
> pub trait MavConnection<M: Message> : MavRawConnection  {
>     fn recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError> {
>         M::parse(self.recv_raw())
>     }
>    ...
> }
> ```
> 
> The issue is that `MAVLinkVXMessageRaw` does not exist and it probably would have to be a trait object.

I still think you'll need a dialect that way since functions like `read_v1_raw_message_async`, etc... Depend on the generic `M`

joaoantoniocardoso:
> > So I had the idea of doing something like that:
> > ```rust
> > pub trait MavRawConnection {
> >     fn recv_raw(&self) -> Result<MAVLinkVXMessageRaw>
> >     ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then have
> > ```rust
> > pub trait MavConnection<M: Message> : MavRawConnection  {
> >     fn recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError> {
> >         M::parse(self.recv_raw())
> >     }
> >    ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The issue is that `MAVLinkVXMessageRaw` does not exist and it probably would have to be a trait object.
> 
> I still think you'll need a dialect that way since functions like `read_v1_raw_message_async`, etc... Depend on the generic `M`
> 
> **P.S: My title might be misleading. My goal is to have Connection wrappers that do not depend on a dialect so you can just receive RAW data**

I think that'd only be possible if we don't check the CRC. For that, we could surely extract the part of the recv raw functions that deal with the message before the CRC, and expose that to the public lib API.

roby2014:
I think it would be great to have some sort of API to just receive/send mavlink messages, even if they are not CRC-validated or from a specific dialect. When the goal is to just "route" messages, I dont want to care about a dialect or version.

joaoantoniocardoso:
> I think it would be great to have some sort of API to just receive/send mavlink messages, even if they are not CRC-validated or from a specific dialect. When the goal is to just "route" messages, I dont want to care about a dialect or version.

I think the main reason the CRC (and thus the knowledge of the message dialect) is necessary for us to create a message is because the protocol relies on a single magic byte to define the start of the message, and this byte can also exist within the rest of the packed, so unless we have another reliable algorithm to identify a message, if we follow the current parser idea, it would possibly discard valid messages while wrongly identifying messages. 

One idea would be not to remove the identified message from the buffer, but to remove the bytes until the next magic byte, which would result in both sending invalid and valid messages, increasing bandwidth. This assumes we can pass the responsibility of validating the messages entirely to the clients.

hamishwillee:
FWIW the design of MAVLink is a known pain-point with routers. In particular the CRC_EXTRA check requires that the router knows about the message - and therefore needs to have all the messages that might be routed to it.. That's annoying because the router doesn't care what the message is but still has to keep up with all new messages.

Further, if a router doesn't have the message definition, the other problem is that it can't determine the target address (since that is embedded in the payload). So it then becomes hard for a router to know where it should send a message. There has been talk of various ways to fix that last one, including pulling the target into the header using an incompatiblility bit.

@tridge Might have some ideas.

onur-ozkan:
Closing this due to inactivity.

PR #292, opened 2024-12-19
Fix: read error leads to message skip for blocking API

kusstas:
ISSUE: Read error leads to message skip. It happens due consuming STX byte of message, even if message parsing isn't completed yet.

GOAL: I have custom readers that could return WouldBlock IO error at any time and code shall interrupt message parsing and return to this later.

### DIFF
--- a/mavlink-core/src/embedded.rs
+++ b/mavlink-core/src/embedded.rs
@@ -5,6 +5,10 @@ const _: () = panic!("Only one of 'embedded' and 'embedded-hal-02' features can 
 
 /// Replacement for std::io::Read + byteorder::ReadBytesExt in no_std envs
 pub trait Read {
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize, MessageReadError> {
+        self.read_exact(buf).map(|_| buf.len())
+    }
+
     fn read_exact(&mut self, buf: &mut [u8]) -> Result<(), MessageReadError>;
 }
 
--- a/mavlink-core/src/error.rs
+++ b/mavlink-core/src/error.rs
@@ -37,6 +37,15 @@ pub enum MessageReadError {
     Parse(ParserError),
 }
 
+impl MessageReadError {
+    pub fn eof() -> Self {
+        #[cfg(feature = "std")]
+        return Self::Io(std::io::ErrorKind::UnexpectedEof.into());
+        #[cfg(any(feature = "embedded", feature = "embedded-hal-02"))]
+        return Self::Io;
+    }
+}
+
 impl Display for MessageReadError {
     fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
         match self {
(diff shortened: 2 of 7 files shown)


pv42:
This seems to be an issue worth fixing. But I think this does not seem to compleatly fix the issue at least as I understand it.
I wrote the following test that uses a reader that alternates beween returning would block and reading a single byte:
```rs
    struct BlockyReader {
        block_next_read: bool,
        index: usize,
    }
    impl Read for BlockyReader {
        fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
            if self.block_next_read {
                self.block_next_read = false;
                Result::Err(Error::new(ErrorKind::WouldBlock, "Test Block"))
            } else {
                let read = HEARTBEAT_V2.get(self.index).ok_or(Error::new(
                    ErrorKind::UnexpectedEof,
                    "EOF",
                ));
                buf[0] = *read?;
                self.index += 1;
                self.block_next_read = true;
                Ok(1)
            }
        }
    }

    #[test]
    fn test_read_error() {
        let mut reader = PeekReader::new(BlockyReader {
            block_next_read: true,
            index: 0,
        });
        loop {
            match read_v2_msg::<mavlink::common::MavMessage, _>(&mut reader) {
                Ok((header, _)) => {
                    assert_eq!(header, crate::test_shared::COMMON_MSG_HEADER);
                    break;
                },
                Err(MessageReadError::Io(err)) if err.kind() == ErrorKind::WouldBlock => (),
                Err(err) => panic!("{err}"),
            }
        }
    }
```
which fails. The problem seems to be that `PeekReader::fetch` calls `io::Read::read_exact()` which then calls `read()` multiple times discarding all previously read data in case of an error.
This PR still fixes the issue when the block occures directly after the STX byte, as in this reader:
```rs
impl Read for BlockyReader {
        fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
            if self.block_next_read {
                self.block_next_read = false;
                Result::Err(Error::new(ErrorKind::WouldBlock, "Test Block"))
            } else {
                let read = HEARTBEAT_V2.get(self.index).ok_or(Error::new(
                    ErrorKind::UnexpectedEof,
                    "EOF",
                ));
                buf[0] = *read?;
                self.index += 1;
                if self.index <= 1 {
                    self.block_next_read = true;
                }
                Ok(1)
            }
        }
    }
```
but if at any other position the message still gets trown out. 
Regardless if this is the intended fix or not there should probably be a test case for it.

kusstas:
Hi, thank you for response. Yeah, I've not tested case where reader can read less data that expect. I've modified `fetch` function to use `read` instead of `read_exact` to store all data that read to buffer and added test cases for each version of mavlink read.

pv42:
[f801bea](https://github.com/mavlink/rust-mavlink/pull/292/commits/f801beaed450543203d1bed23447e28c01dc1342) seems to have broken the `process_log_files` test, it does not terminate anymore (at least for me). 
This seems to be happening because `std::io::default_read_exact()` that performs the work for `read_exact()` has a special abort condition where it returns an error when 0 bytes where read:
```rs
while !buf.is_empty() {
    match this.read(buf) {
        Ok(0) => break,
        ...
    }
}
if !buf.is_empty() { Err(Error::READ_EXACT_EOF) } else { Ok(()) }
```

kusstas:
Yeah, I've added case for zero check on read, and now seems tests with features listed in workflow work well.

patrickelectric (approved):

Issue #285, opened 2024-09-24
Fix codegeneration for newer MAVLink message definitions


pv42:
Currently using `mavlink` against [mavlink](https://github.com/mavlink/mavlink/) master's message definitions does not compile.
There are 2 issues, both related to bitmasks:
- The enum `GIMBAL_MANAGER_FLAGS` that is `bitmask="true"` is used in messages without labeling the fields as `display="bitmask"`. I believe this is a bug and have opened a PR (https://github.com/mavlink/mavlink/pull/2154)
- The enum `ILLUMINATOR_MODE` that is not a `bitmask` is used with `display="bitmask"` in the `ILLUMINATOR_STATUS` message. This is not really a bug since `ILLUMINATOR_MODE` is both used as an bitmask to indicate supported modes as well as a value to indicate the current mode. 

At least the second problem requires a solution that allows the usage of enums as bitmasks independent of their designation in the enum definition.
This could be solved by generating a bitmask version of every enum regardless of their `bitmask="true"`designation but that would probably create a lot of extra, unneeded code. The solution that avoids that by checking the usage of each enum and the generate it depending on how it is used also does not seem ideal to me since the code generation would probably not be context free anymore. This could impact code generation performance and increase complexity but I haven't looked too far into it. 
Since the version of the mavlink message definitons used is now over a year old a solution should be found, so newer messages/message fields can be used. 

If someone needs an immediate solution that does not require the `ILLUMINATOR_STATUS` message, it would also be possible to just blacklist this message from code generation until a more permanent solution is found.

PR #302, opened 2025-05-02
Update MAVLINK message definitions

chemicstry:
Current MAVLINK message definitions are 2 years old and many new messages are missing.


chemicstry:
There seems to be some problems with the embedded build, marking as draft until I figure it out

chemicstry:
I think it should be OK now. The `display="bitfield"` was deprecated in MAVLINK and now it is defined as `bitmask="true"` on enum instead. However, there are still cases of both so we have to support each case.

pv42:
Last time I looked into updating MAVLink definition the issue was that there are enums that are both used as bitmasks and as single choice selections (See #285) but error message looks different this time.

hamishwillee:
Are you actually using the mavlink version for anything? i.e. does your code check it? AFAIK no code bothers to check this so there is no point updating it. If we start a release process, then perhaps we will.

I will look at the build issues tomorrow.

chemicstry:
Sorry for the confusing PR name, I meant to update message definitions.

patrickelectric:
@chemicstry can you rebase it ?

patrickelectric:
@chemicstry can you rebase the code running `git base origin/master` ? It appears that still have conflicts, I did try using github web tool but it's still failing.

chemicstry:
I opened #308 with clean git history

PR #320, opened 2025-06-11
implement `try_recv` to `MavConnection`

onur-ozkan:
Right now, in order to release the `recv` lock in my current project, I have to do this:

```rs
async fn message_loop(self: Arc<Self>) {
    loop {
        let recv_future = smol::unblock({
            let _self = self.clone();
            move || _self.connection.recv()
        });

        match recv_future.timeout(Duration::from_millis(200)).await {
            None => {
                simple_log::warning!("Timed out while waiting for message!");
                Timer::after(Duration::from_millis(200)).await;
            },
            // Rest of the work
        }
    }
}
```

which is problematic because:

1. It requires a lot of clones.
2. I have to convert `recv()` into a `Future`.
3. I have to wrap it with a timeout.
4. Points 2 and 3 require additional dependencies.
5. This approach is bad for performance on low-power devices.

Instead, I would prefer to simply do this:

```rs
async fn message_loop(&self) {
    loop {
        match self.connection.try_recv() {
            None => {
                // Prevent busy-looping
                Timer::after(Duration::from_millis(50)).await;
            },
            // Rest of the work
        }
    }
}
```

which solves all the issues mentioned above.

This patch implements `try_recv` for the MavConnection trait, which allows users (like me) to have more control over their message receiving implementations. I looked at `AsyncMavConnection`, but as far as I can see it's non-blocking anyway. I am not sure why but it doesn't seem consistent with the `MavConnection` trait.

### DIFF
--- a/mavlink-core/src/connection/tcp.rs
+++ b/mavlink-core/src/connection/tcp.rs
@@ -96,6 +96,10 @@ impl<M: Message> MavConnection<M> for TcpConnection {
         result
     }
 
+    fn try_recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError> {
+        self.recv()
+    }
+
     fn send(&self, header: &MavHeader, data: &M) -> Result<usize, crate::error::MessageWriteError> {
         let mut lock = self.writer.lock().unwrap();
 
--- a/mavlink-core/src/connection/direct_serial.rs
+++ b/mavlink-core/src/connection/direct_serial.rs
@@ -52,6 +52,20 @@ impl<M: Message> MavConnection<M> for SerialConnection {
         }
     }
 
+    fn try_recv(&self) -> Result<(MavHeader, M), MessageReadError> {
+        let mut port = self.port.lock().unwrap();
+        let version = ReadVersion::from_conn_cfg::<_, M>(self);
+
+        #[cfg(not(feature = "signing"))]
+        let result = read_versioned_msg(port.deref_mut(), version);
+
+        #[cfg(feature = "signing")]
+        let result =
+            read_versioned_msg_signed(port.deref_mut(), version, self.signing_data.as_ref());
+
+        result
+    }
+
     fn send(&self, header: &MavHeader, data: &M) -> Result<usize, MessageWriteError> {
         let mut port = self.port.lock().unwrap();
 
--- a/mavlink-core/src/connection/file.rs
+++ b/mavlink-core/src/connection/file.rs
@@ -64,6 +64,19 @@ impl<M: Message> MavConnection<M> for FileConnection {
         }
     }
 
+    fn try_recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError> {
+        let mut file = self.file.lock().unwrap();
+        let version = ReadVersion::from_conn_cfg::<_, M>(self);
+
+        #[cfg(not(feature = "signing"))]
+        let result = read_versioned_msg(file.deref_mut(), version);
+        #[cfg(feature = "signing")]
+        let result =
+            read_versioned_msg_signed(file.deref_mut(), version, self.signing_data.as_ref());
+
+        result
+    }
+
     fn send(&self, _header: &MavHeader, _data: &M) -> Result<usize, MessageWriteError> {
         Ok(0)
     }
--- a/mavlink-core/src/connection/mod.rs
+++ b/mavlink-core/src/connection/mod.rs
@@ -24,6 +24,12 @@ pub trait MavConnection<M: Message> {
     /// Blocks until a valid frame is received, ignoring invalid messages.
     fn recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError>;
 
+    /// Try to receive a MAVLink message.
+    ///
+    /// Non-blocking variant of `recv()`, returns immediately with a `MessageReadError`
+    /// if there is an error or no message is available.
+    fn try_recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError>;
+
     /// Send a MAVLink message
     fn send(&self, header: &MavHeader, data: &M) -> Result<usize, crate::error::MessageWriteError>;
 
--- a/mavlink-core/src/connection/udp.rs
+++ b/mavlink-core/src/connection/udp.rs
@@ -102,6 +102,25 @@ impl<M: Message> MavConnection<M> for UdpConnection {
         }
     }
 
+    fn try_recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError> {
+        let mut reader = self.reader.lock().unwrap();
+        let version = ReadVersion::from_conn_cfg::<_, M>(self);
+
+        #[cfg(not(feature = "signing"))]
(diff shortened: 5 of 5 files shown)


mavlink-core/src/connection/tcp.rs
```diff
@@ -96,6 +96,11 @@ impl<M: Message> MavConnection<M> for TcpConnection {
         result
     }
 
+    fn try_recv(&self) -> Result<Option<(MavHeader, M)>, crate::error::MessageReadError> {
+        // `TcpConnection::recv` doesn't seem to block?
+        self.recv().map(Some)
+    }
```
onur-ozkan: I don't know what to do here. Alternatively I can put `unimplemented!()` but I am not sure if that would be a good idea.

joaoantoniocardoso: It is different from the others because the TCP buffer is already framed, so it is assumed that the received buffer is a complete message.

onur-ozkan:
> Hi, I see that the Option is very simple and solves your case here, but masking parsing errors as `None` might narrow its use. For instance, if someone is trying to build a small router, parsing errors might be instrumental.
> 
> So maybe we could have something around
> 
> ```rust
> enum TryRecvError {
>     Empty,
>     Disconnected,
>     Parse(ParserError),
> }
> ```
> 
> What do you think?

We can return the inner `Err` directly without creating another error type. This patch should work for everyone (I think):

```diff
diff --git a/mavlink-core/src/connection/direct_serial.rs b/mavlink-core/src/connection/direct_serial.rs
index bb0fefe..f9f2a7a 100644
--- a/mavlink-core/src/connection/direct_serial.rs
+++ b/mavlink-core/src/connection/direct_serial.rs
@@ -52,7 +52,7 @@ impl<M: Message> MavConnection<M> for SerialConnection {
         }
     }
 
-    fn try_recv(&self) -> Result<Option<(MavHeader, M)>, MessageReadError> {
+    fn try_recv(&self) -> Result<(MavHeader, M), MessageReadError> {
         let mut port = self.port.lock().unwrap();
         let version = ReadVersion::from_conn_cfg::<_, M>(self);
 
@@ -63,13 +63,7 @@ impl<M: Message> MavConnection<M> for SerialConnection {
         let result =
             read_versioned_msg_signed(port.deref_mut(), version, self.signing_data.as_ref());
 
-        match result {
-            Ok(msg) => Ok(Some(msg)),
-            Err(MessageReadError::Io(e)) if e.kind() == io::ErrorKind::UnexpectedEof => {
-                Err(MessageReadError::Io(e))
-            }
-            Err(_) => Ok(None),
-        }
+        result
     }
 
     fn send(&self, header: &MavHeader, data: &M) -> Result<usize, MessageWriteError> {
diff --git a/mavlink-core/src/connection/file.rs b/mavlink-core/src/connection/file.rs
index 61adb81..055fd2e 100644
--- a/mavlink-core/src/connection/file.rs
+++ b/mavlink-core/src/connection/file.rs
@@ -64,7 +64,7 @@ impl<M: Message> MavConnection<M> for FileConnection {
         }
     }
 
-    fn try_recv(&self) -> Result<Option<(MavHeader, M)>, crate::error::MessageReadError> {
+    fn try_recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError> {
         let mut file = self.file.lock().unwrap();
         let version = ReadVersion::from_conn_cfg::<_, M>(self);
 
@@ -74,13 +74,7 @@ impl<M: Message> MavConnection<M> for FileConnection {
         let result =
             read_versioned_msg_signed(file.deref_mut(), version, self.signing_data.as_ref());
 
-        match result {
-            Ok(msg) => Ok(Some(msg)),
-            Err(MessageReadError::Io(e)) if e.kind() == io::ErrorKind::UnexpectedEof => {
-                Err(MessageReadError::Io(e))
-            }
-            Err(_) => Ok(None),
-        }
+        result
     }
 
     fn send(&self, _header: &MavHeader, _data: &M) -> Result<usize, MessageWriteError> {
diff --git a/mavlink-core/src/connection/mod.rs b/mavlink-core/src/connection/mod.rs
index 1ac4da5..d6435cd 100644
--- a/mavlink-core/src/connection/mod.rs
+++ b/mavlink-core/src/connection/mod.rs
@@ -27,7 +27,7 @@ pub trait MavConnection<M: Message> {
     /// Try to receive a MAVLink message.
     ///
     /// Non-blocking variant of `recv()`, should return `None` immediately if no message is available.
-    fn try_recv(&self) -> Result<Option<(MavHeader, M)>, crate::error::MessageReadError>;
+    fn try_recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError>;
 
     /// Send a MAVLink message
     fn send(&self, header: &MavHeader, data: &M) -> Result<usize, crate::error::MessageWriteError>;
diff --git a/mavlink-core/src/connection/tcp.rs b/mavlink-core/src/connection/tcp.rs
index 76d535b..d1dc4df 100644
--- a/mavlink-core/src/connection/tcp.rs
+++ b/mavlink-core/src/connection/tcp.rs
@@ -96,9 +96,8 @@ impl<M: Message> MavConnection<M> for TcpConnection {
         result
     }
 
-    fn try_recv(&self) -> Result<Option<(MavHeader, M)>, crate::error::MessageReadError> {
-        // `TcpConnection::recv` doesn't seem to block?
-        self.recv().map(Some)
+    fn try_recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError> {
+        self.recv()
     }
 
     fn send(&self, header: &MavHeader, data: &M) -> Result<usize, crate::error::MessageWriteError> {
diff --git a/mavlink-core/src/connection/udp.rs b/mavlink-core/src/connection/udp.rs
index b808b2f..aab993d 100644
--- a/mavlink-core/src/connection/udp.rs
+++ b/mavlink-core/src/connection/udp.rs
@@ -102,7 +102,7 @@ impl<M: Message> MavConnection<M> for UdpConnection {
         }
     }
 
-    fn try_recv(&self) -> Result<Option<(MavHeader, M)>, crate::error::MessageReadError> {
+    fn try_recv(&self) -> Result<(MavHeader, M), crate::error::MessageReadError> {
         let mut reader = self.reader.lock().unwrap();
         let version = ReadVersion::from_conn_cfg::<_, M>(self);
 
@@ -118,7 +118,7 @@ impl<M: Message> MavConnection<M> for UdpConnection {
             }
         }
 
-        Ok(result.ok())
+        result
     }
 
     fn send(&self, header: &MavHeader, data: &M) -> Result<usize, crate::error::MessageWriteError> {
```

What do you think?

joaoantoniocardoso:
That is fine, too. By doing that, we are leaving the responsibility of the semantics to the user -- he will need to figure out what error to map the his reconnection logic, for example.

Just be sure to update the trait comments :)

onur-ozkan:
Thank you, I will update the comments tomorrow or the day after.

onur-ozkan:
> Thank you, I will update the comments tomorrow or the day after.

Done :)

patrickelectric (approved):

PR #325, opened 2025-06-19
remove outdated `serial` and migrate to `serialport`

onur-ozkan:
Last activity on [`serial-rs`](https://github.com/dcuddeback/serial-rs) was 7 years ago and it has not been maintained since then, which can be a big concern on production-grade projects.

Other bumps aims to remove the duplicated dependencies introduced by the `serialport` crate. Also, more importantly, to use a more up-to-date dependency version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ resolver = "1"
 crc-any = { version = "2.3.5", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 num-derive = "0.4"
-bitflags = "1.2.1"
+bitflags = { version = "2.9.1", default-features = false }
 byteorder = { version = "1.3.4", default-features = false }
 
 [workspace.package]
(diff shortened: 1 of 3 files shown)


mavlink-core/src/connectable.rs
```diff
@@ -42,12 +42,12 @@ impl Display for UdpConnectable {
...
 pub struct SerialConnectable {
     pub(crate) port_name: String,
-    pub(crate) baud_rate: usize,
+    pub(crate) baud_rate: u32,
 }
 
 impl SerialConnectable {
     /// Creates a serial connection address with port name and baud rate.
-    pub fn new(port_name: String, baud_rate: usize) -> Self {
+    pub fn new(port_name: String, baud_rate: u32) -> Self {
```
joaoantoniocardoso, line 50: @patrickelectric Just a reminder that this is a breaking change, next release should be greater than or equal to `0.15`.

joaoantoniocardoso (approved):

PR #329, opened 2025-06-20
refactor: split serial port into separate r/w mutexes

### LINKED ISSUE #326: `MavConnection::recv` can cause `MavConnection::send` to hang even on different threads
We acquire the lock at the beginning of `MavConnection::send`. Which means, in the loop, if we repeatedly hit `_` part of the `match` (which is any error that isn't `ErrorKind::UnexpectedEof`) here
 https://github.com/mavlink/rust-mavlink/blob/7123184db4bcadb2f2ba88086fbf288a0d2ee1fe/mavlink-core/src/connection/direct_serial.rs#L41-L51

`MavConnection::send` will hang while trying to lock the port since `recv` never releases it. This happened to me couple times and I thought I have issues on my thread pool.

Ideally, we should return `io::ErrorKind::TimedOut` if no message is received within a certain time interval (e.g., 5 seconds).

Any opinions? @patrickelectric

onur-ozkan:
Use separate port instances for reading and writing operations to prevent deadlock when concurrent `recv()` and `send()` calls compete for the same mutex.

Resolves #326

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,7 +9,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
         ref: ${{ github.event.repository.default_branch }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ jobs:
   formatting:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt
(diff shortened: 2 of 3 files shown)


onur-ozkan:
I have no idea why CI is trying to build on the old codebase and fail..?

onur-ozkan:
I [checked](https://github.com/mavlink/rust-mavlink/actions/caches) to see if we have any cache, but there is no cache. It's really strange.

onur-ozkan:
image

I wonder if we hit some GHA bug or something :D

onur-ozkan:
I will try to create another PR to see if it fails there too.

onur-ozkan:
> I will try to create another PR to see if it fails there too.

It didn't work: https://github.com/mavlink/rust-mavlink/pull/330.

> Yeah, I saw that and it makes no sense to me either. I'm having trouble with my internet connection today, but could you try running the tests on your machine just to be sure? I think we can move forward in this case.

I tried many of them and all worked fine on clean checkout locally.

onur-ozkan:
Are we using custom/self-hosted CI runners? Maybe that could be related.

onur-ozkan:
> Ok figured it out: this PR was based on master before `try_receive` was merged, this PR does not change the `port` used in `try_recv`.
> 
> https://github.com/mavlink/rust-mavlink/blob/efb6fed0748334306a2e36012230891d9b1dc34c/mavlink-core/src/connection/direct_serial.rs#L59

I just figured that out too, that PR got us all lol :smile:

PR #332, opened 2025-06-20
refactor: split serial port into separate r/w mutexes

### LINKED ISSUE #326: `MavConnection::recv` can cause `MavConnection::send` to hang even on different threads

onur-ozkan:
Use separate port instances for reading and writing operations to prevent deadlock when concurrent `recv()` and `send()` calls compete for the same mutex.

Resolves #326

### DIFF
--- a/mavlink-core/src/connection/direct_serial.rs
+++ b/mavlink-core/src/connection/direct_serial.rs
@@ -20,7 +20,10 @@ use crate::{read_versioned_msg_signed, write_versioned_msg_signed, SigningConfig
 use super::Connectable;
 
 pub struct SerialConnection {
-    port: Mutex<PeekReader<Box<dyn SerialPort>>>,
+    // Separate ports for reading and writing as it's safe to use concurrently.
+    // See the official ref: https://github.com/serialport/serialport-rs/blob/321f85e1886eaa1302aef8a600a631bc1c88703a/examples/duplex.rs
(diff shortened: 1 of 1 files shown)


pv42:
The issue seems that port in try_recv was not adjusted
https://github.com/mavlink/rust-mavlink/blob/master/mavlink-core/src/connection/direct_serial.rs#L56

onur-ozkan:
This was previously approved, so I will re-merge it once CI is green.

I think it's worth doing a `0.15` release since we did several improvements, fixes and new API additions. I am not familiar with our release process yet, so I will leave that part to you if you don't mind @patrickelectric.

Issue #321, opened 2025-06-11
async implementation re-design


onur-ozkan:
If we can make `AsyncMavConnection` tokio-free and more generic, that would allow anyone with any async runtime (especially people who build on low-power devices) to use this trait, which would be amazing.

onur-ozkan:
Status quo: the work is nearly done on the serialport side, it's awaiting for reviews and feedbacks.

PR #341, opened 2025-06-30
use tokio-free locks and results

onur-ozkan:
Aims to reduce the dependency on tokio in async mode. This is the first step of https://github.com/mavlink/rust-mavlink/issues/321 and the ultimate goal is to remove tokio entirely from the async interface.

Helps to https://github.com/mavlink/rust-mavlink/issues/321.

### DIFF
--- a/mavlink-core/Cargo.toml
+++ b/mavlink-core/Cargo.toml
@@ -17,22 +17,22 @@ edition.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-crc-any = { workspace = true, default-features = false }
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

joaoantoniocardoso:
> I didn't change the API yet, I am planning to change it only when #321 is entirely closed. Then I can replace `tokio-1` feature with `async` or something.

I assumed it changed since the error type changed from `tokio::io::ErrorKind` to `std::io::ErrorKind` in the `mavlink-core/src/async_peek_reader.rs` file.

PR #343, opened 2025-07-01
String parsing for char[]

joaoantoniocardoso:
> This allows messages like STATUSTEXT to parse nicely as an ArrayString in Rust.

I'm taking over #193 so we can implement the feature more quickly.


gbin:
Thanks @joaoantoniocardoso

onur-ozkan (approved):
LGTM

onur-ozkan:
Should we revert this? It didn't do anything better as explained in https://github.com/mavlink/rust-mavlink/pull/345.

PR #349, opened 2025-07-16
Make `remove_trailing_zeroes` panic-safe and more efficient

onur-ozkan:
Refactors `remove_trailing_zeroes` to use a simple `while` loop instead of iterating over a reversed slice. This makes the code clearer and more efficient by directly checking bytes from the end while still ensuring at least one byte remains.

Additionally, previous version was not panic-safe and would panic on empty inputs (like the `test_remove_trailing_zeroes_empty_slice` test). This change also fixes that and makes the function panic-safe.

### DIFF
--- a/mavlink-core/src/utils.rs
+++ b/mavlink-core/src/utils.rs
@@ -9,11 +9,7 @@
 pub fn remove_trailing_zeroes(data: &[u8]) -> usize {
     let mut len = data.len();
 
-    for b in data[1..].iter().rev() {
-        if *b != 0 {
-            break;
-        }
-
+    while len > 1 && data[len - 1] == 0 {
         len -= 1;
     }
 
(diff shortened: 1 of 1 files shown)


mavlink-core/src/utils.rs
```diff
@@ -114,3 +110,13 @@ impl RustDefault for f64 {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_remove_trailing_zeroes_empty_slice() {
+        remove_trailing_zeroes(&[0]);
```
joaoantoniocardoso, line 120: pretty sure this should be `remove_trailing_zeroes(&[]);` instead.

onur-ozkan: Yeah, I forgot to remove `0` after debugging with `0` :)

PR #353, opened 2025-07-28
mavlink-bindgen: skip non XML files

azerupi:
When running mavlink-bindgen in a directory that contains other files it fails. In my case it panicked when trying to "generate" bindings for `.git`. 

**Example directory**
```
mavlink
├── .git
├── all.xml
├── custom.xml
├── common.xml
├── minimal.xml
├── README.md
└── standard.xml
```

This PR adds a check when iterating over the entries in the directory to skip hidden files, directories and non-XML files.

### DIFF
--- a/mavlink-bindgen/src/lib.rs
+++ b/mavlink-bindgen/src/lib.rs
@@ -55,6 +55,11 @@ fn _generate(
         let definition_file = PathBuf::from(entry.file_name());
         let module_name = util::to_module_name(&definition_file);
 
+        // Skip non-XML files
+        if !definition_file.extension().is_some_and(|e| e == "xml") {
+            continue;
+        }
+
         let definition_rs = PathBuf::from(&module_name).with_extension("rs");
 
         let dest_path = destination_dir.join(definition_rs);


mavlink-bindgen/src/lib.rs
```diff
@@ -52,6 +52,14 @@ fn _generate(
             }
         })?;
 
+        // Skip hidden files, directories, and non-XML files
+        if entry.file_name().to_string_lossy().starts_with(".")
+            || !entry.file_type().unwrap().is_file()
+            || entry.path().extension().unwrap() != "xml"
```
onur-ozkan: "non-XML" should be a sufficient enough check.

onur-ozkan: "xml" check can also be in a panic-safe way, e.g., `entry.file_name().ends_with("xml")`

azerupi: Indeed, you are right.  :+1: 
I've made the changes as you proposed with one small change, `file_name()` gives an `OsString` instead of a `str` which doesn't have `ends_with` so I compare the bytes.

onur-ozkan: Right, you could move 1 line below and do that on `definition_file`.

azerupi: Yes but the `ends_with` on paths works differently. From the docs it says that it only matches on components and there is an example showing that checking the extension through it doesn't work.

```rust
assert_eq!(false, Path::new("test.xml").ends_with("xml"));
```

But I think I found something elegant that works equally well.

mavlink-bindgen/src/lib.rs
```diff
@@ -55,6 +55,11 @@ fn _generate(
         let definition_file = PathBuf::from(entry.file_name());
         let module_name = util::to_module_name(&definition_file);
 
+        // Skip non-XML files
+        if definition_file.extension().is_some_and(|e| e == "xml") {
```
onur-ozkan: Wouldn't this skip all the XML files?

azerupi: Indeed, I'm sorry about that. I forgot to patch the dependency to my local version instead of this branch when testing which made me believe this was working.

onur-ozkan: It can happen, no worries. Thanks for the effort. :)

onur-ozkan (approved):

mavlink-bindgen/src/lib.rs
azerupi: Thank you for your patience! :)

Issue #347, opened 2025-07-05
limitations of dynamic connection dispatching


onur-ozkan:
I am working on adding read and write buffering for serial connections. But there is a problem: there is no clear way to handle connection-type specific configurations (e.g., adding options specifically for `SerialConnection` and exposing them to users). This is because the way connection is initialized is too generic and it's somewhat magical with hiding the actual connection type.

In action, this is what users do today to initialize connections and do some configuration:

```rust
let mut mavconn = mavlink::connect::<mavlink::ardupilotmega::MavMessage>(...);
mavconn.set_protocol_version(mavlink::MavlinkVersion::V1);
```

However, this approach can't work for every case. When connection parameters become configurable, setting them after initialization doesn't always make sense because the connection type is already settled.

Instead of this dynamic connection dispatch, I suggest moving to an explicit connection initialization approach. So then:

1. We don't have to adapt every connection type just because we want to add configuration for a specific one.
2. We can expose connection-specific configuration types that don't need to be generic or compatible across all connection types.

By roughly thinking, something like this would be simple and efficient:

```rust
// Using `SerialConnection` for demonstration purposes, this applies to each connection type.
let mut config = SerialConnectionConfig::new();
config.disable_x();
config.enable_y();
config.set_z(5);

let con = SerialConnection::connect(or new)(config);
```

cc @patrickelectric @joaoantoniocardoso @pv42

onur-ozkan:
> There are the already the implementations of the Connectable trait e.g. SerialConnectable. They currently just represent a parsed address string but adding more options there would be possible.

I am aware of that but it didn't look like a decent approach to add new fields there and then propagate them. To me, I think what we should do is just make the connection type more explicit like I proposed in the issue, or even something like this could help:


```rust
enum ConnectionConfiguration {
  Serial(SerialConnectionConfig),
  File(FileConnectionConfig),
  ...
}


// Update `mavlink::connect` function to take `ConnectionConfiguration`.
mavlink::connect(ConnectionConfiguration::Serial(config));
```

pv42:
I am personally in favor of at least having a `connect_str(addr: &str) -> ...` function exposed since it would allow to add the mavlink connection with only a single string parameter compared to the rather complex `ConnectionConfiguration`. This does not neccesary have to dyn it could also an `enum`. Alternatively at least something like `parse_address(str).connect()` should be present.

[...]

PR #355, opened 2025-08-05
implement config layer for connection types

onur-ozkan:
Implements configuration layer for each connection type without breaking the existing `mavlink::connect` API, so it can still be used as-is.

In short, this allows users to make connections with specific configuration types.

e.g.,: A serial connection (and all other types) can now be established like this:

```rust
use mavlink::{Connectable, SerialConfig};

let mut config = SerialConfig::new("/dev/ttyTHS1".to_owned(), 115200);
// modify the configuration
config.connect::<mavlink::ardupilotmega::MavMessage>();
```

See [limitations of dynamic connection dispatching #347](https://github.com/mavlink/rust-mavlink/issues/347) for the motivation of this PR.

Resolves: https://github.com/mavlink/rust-mavlink/issues/347


mavlink-core/src/async_connection/direct_serial.rs
```diff
@@ -113,7 +111,7 @@ impl<M: Message + Sync + Send> AsyncMavConnection<M> for AsyncSerialConnection {
 }
 
 #[async_trait]
-impl AsyncConnectable for SerialConnectable {
```
pv42, line 116: I think the trait should be renamed too, same for the none-async version.

onur-ozkan: I think the trait name is fine and doesn't need to be related to config types. Not to mention changing it would break things for all users.

pv42: Ok fair point :+1:

joaoantoniocardoso (approved):
I'm okay with it, thanks!

PR #357, opened 2025-08-11
Add a method to retrieve the target system/component if it exists

azerupi:
A good amount of MAVLink messages contain a target system ID and target component ID. It is a bit unfortunate that this is only a convention in the payload and is not more standardized.

For my use case I need to filter out the messages that are not addressed to my system and this adds a convenient way to achieve this. Prior to this change I needed to create a giant match statement to map messages containing those fields manually.

Do I need to add tests for this? I didn't find an obvious example that I could copy.

Generated code looks like this:
```rust
fn target_system_id(&self) -> Option<u8> {
        match self {
            Self::LOG_REQUEST_DATA(inner) => Some(inner.target_system),
            Self::PING(inner) => Some(inner.target_system),
            Self::COMMAND_CANCEL(inner) => Some(inner.target_system),
            Self::PARAM_SET(inner) => Some(inner.target_system),
            // many more ...
            _ => None,
        }
}
```

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -1714,3 +1714,71 @@ fn capitalize_word(text: impl AsRef<str>) -> String {
         Some(char) => char.to_uppercase().to_string() + &chars.as_str().to_ascii_lowercase(),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn emits_target_id_match_arms() {
+        // Build a minimal profile containing one message with target fields and one without
+        let mut profile = MavProfile::default();
+
+        let msg_with_targets = MavMessage {
+            id: 300,
+            name: "COMMAND_INT".to_string(),
+            description: None,
+            fields: vec![
+                MavField {
+                    mavtype: MavType::UInt8,
+                    name: "target_system".to_string(),
+                    description: None,
+                    enumtype: None,
+                    display: None,
+                    is_extension: false,
+                },
+                MavField {
+                    mavtype: MavType::UInt8,
+                    name: "target_component".to_string(),
+                    description: None,
+                    enumtype: None,
+                    display: None,
+                    is_extension: false,
+                },
+            ],
+        };
+
+        let msg_without_targets = MavMessage {
+            id: 0,
+            name: "HEARTBEAT".to_string(),
+            description: None,
+            fields: vec![MavField {
+                mavtype: MavType::UInt32,
+                name: "custom_mode".to_string(),
+                description: None,
+                enumtype: None,
+                display: None,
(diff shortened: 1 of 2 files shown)


mavlink-bindgen/src/parser.rs
```diff
@@ -333,6 +337,48 @@ impl MavProfile {
             }
         }
     }
+
+    fn emit_mav_message_target_system_id(&self) -> TokenStream {
+        let arms: Vec<TokenStream> = self
+            .messages
+            .values()
+            .filter(|msg| msg.fields.iter().any(|f| f.name == "target_system"))
```
onur-ozkan, line 345: Do we ever have more than one `target_system` field in a single message? If not, we should panic in that case.

cc @patrickelectric @joaoantoniocardoso

onur-ozkan (commented):
> Do I need to add tests for this? I didn't find an obvious example that I could copy.

Yes, for any addition we should have a coverage. You can find plenty of `MavMessage` references in `mavlink/tests`.

onur-ozkan (approved):
> I have added test in the way you proposed to check that specific messages from the common spec return the right values. I can add more messages to test if desired.
> 
> But I thought it was a bit light in terms of testing so I added a test to the code generator as well. There weren't any other tests in there and in the long run there might be a better way to run tests on the generator by actually having a "test dialect" with xml files and checking the generated code. So if you want me to remove that test, let me know.

It's fine we can keep them and I agree, we should have more coverage on the generator with dummy xml files in the future.

Changes looks good to me, I will hold off on merging until we get a response to [this comment](https://github.com/mavlink/rust-mavlink/pull/357#discussion_r2272433890).

mavlink-bindgen/src/parser.rs
azerupi: I don't think it is valid for any field to be duplicated, is it?
The generated code would fail to build because two fields on the `_DATA` struct would have the same name.

onur-ozkan: Fair, then we should `assert` it (that it only finds single `target_system` in `msg.fields`) to make it future-proof as the generated code might not be covered by CI for some reason and we could end up shipping broken generated code.

azerupi: Ok, I think it makes more sense to validate this for all fields. I can submit a follow-up PR with these changes: https://github.com/azerupi/rust-mavlink/pull/1 I'll change the target to this repo after this PR is merged if that is ok for you.

onur-ozkan: Looks good to me. :)

azerupi: Alright, shall we merge this PR then? :slightly_smiling_face:

PR #359, opened 2025-08-14
Restore old bitflags serialization behavior.

### LINKED ISSUE #348: 0.15.0 bitflags serialization
Serialization of bitflags has changed from 0.14.1 to 0.15.0. Is this an intended change or a bug?

0.15.0 (Serialization of HEARTBEAT base_mode is a string: 
```
"MAV_MODE_FLAG_SAFETY_ARMED | MAV_MODE_FLAG_MANUAL_INPUT_ENABLED | MAV_MODE_FLAG_STABILIZE_ENABLED | MAV_MODE_FLAG_GUIDED_ENABLED | MAV_MODE_FLAG_CUSTOM_MODE_ENABLED"
```

0.14.1 (Serialization of HEARTBEAT base_mode is an object with bits as number:
```
{ bits: 217 }
```

Another thing I noticed: with  "format-generated-code" feature enabled the bitflags definition are not "pretty" formatted, for example:
```
bitflags! { # [cfg_attr (feature = "serde" , derive (Serialize , Deserialize))] # [cfg_attr (feature = "arbitrary" , derive (Arbitrary))] # [derive (Debug , Copy , Clone , PartialEq)] # [doc = "These flags encode the MAV mode."] pub struct MavModeFlag : u8 { # [doc = "0b10000000 MAV safety set to armed. Motors are enabled / running / can start. Ready to fly. Additional note: this flag is to be ignore when sent in the command MAV_CMD_DO_SET_MODE and MAV_CMD_COMPONENT_ARM_DISARM shall be used instead. The flag can still be used to report the armed state."] const MAV_MODE_FLAG_SAFETY_ARMED = 128 ; # [doc = "0b01000000 remote control input is enabled."] const MAV_MODE_FLAG_MANUAL_INPUT_ENABLED = 64 ; # [doc = "0b00100000 hardware in the loop simulation. All motors / actuators are blocked, but internal software is full operational."] const MAV_MODE_FLAG_HIL_ENABLED = 32 ; # [doc = "0b00010000 system stabilizes electronically its attitude (and optionally position). It needs however further control inputs to move around."] const MAV_MODE_FLAG_STABILIZE_ENABLED = 16 ; # [doc = "0b00001000 guided mode enabled, system flies waypoints / mission items."] const MAV_MODE_FLAG_GUIDED_ENABLED = 8 ; # [doc = "0b00000100 autonomous mode enabled, system finds its own goal positions. Guided flag can be set or not, depends on the actual implementation."] const MAV_MODE_FLAG_AUTO_ENABLED = 4 ; # [doc = "0b00000010 system has a test mode enabled. This flag is intended for temporary system tests and should not be used for stable implementations."] const MAV_MODE_FLAG_TEST_ENABLED = 2 ; # [doc = "0b00000001 Reserved for future use."] const MAV_MODE_FLAG_CUSTOM_MODE_ENABLED = 1 ; } }
```

pv42:
Fixes #348:
- add custom (`de`)`serialize` implementations for bitflags fields as suggested in [bitflags-serde-legacy](https://github.com/KodrAus/bitflags-serde-legacy)

Tests:
- add two `serde` test
- one for `common::MavMessage::HEARTBEAT` that cover enums and bitflags
- one for `test::MavMessage::TEST_TYPES` that covers all other data types
- this uses the no longer maintained [serde_test](https://crates.io/crates/serde_test) crate, but I think this is still superior to writing custom code
- the test assume a fixed field order in the message serialization

### DIFF
--- a/mavlink/Cargo.toml
+++ b/mavlink/Cargo.toml
@@ -31,6 +31,7 @@ mavlink-core = { version="=0.15.0", path = "../mavlink-core", default-features =
 num-traits = { workspace = true, default-features = false }
 num-derive = { workspace = true }
 bitflags = { workspace = true }
+bitflags-serde-legacy = { version = "0.1.1", optional = true}
 serde = { version = "1.0.115", optional = true, features = ["derive"] }
 serde_arrays = { version = "0.2.0", optional = true }
 arbitrary = { version = "1.4", optional = true, features = ["derive"] }
@@ -94,7 +95,7 @@ direct-serial = ["mavlink-core/direct-serial"]
 # Use 'embedded-hal-0.2' feature to enable deprecated embedded-hal=0.2.3 (some hals is not supports embedded-hal=1.0 yet).
 embedded = ["mavlink-core/embedded"]
 embedded-hal-02 = ["mavlink-core/embedded-hal-02"]
-serde = ["bitflags/serde", "dep:serde", "dep:serde_arrays", "mavlink-core/serde", "arrayvec/serde"]
+serde = ["bitflags/serde", "dep:serde", "dep:serde_arrays", "mavlink-core/serde", "arrayvec/serde", "dep:bitflags-serde-legacy"]
 tokio-1 = ["mavlink-core/tokio-1"]
 arbitrary = ["dep:arbitrary", "dep:rand", "mavlink-bindgen/arbitrary", "mavlink-core/arbitrary"]
 
--- a/mavlink/Cargo.toml
+++ b/mavlink/Cargo.toml
@@ -115,3 +115,4 @@ features = [
 
 [dev-dependencies]
 tokio = { version = "1.0", default-features = false, features = ["macros", "rt", "time" ] }
+serde_test = "1.0"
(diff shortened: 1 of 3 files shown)


mavlink/Cargo.toml
```diff
@@ -31,6 +31,7 @@ mavlink-core = { version="=0.15.0", path = "../mavlink-core", default-features =
 num-traits = { workspace = true, default-features = false }
 num-derive = { workspace = true }
 bitflags = { workspace = true }
+bitflags-serde-legacy = { version = "0.1.1", optional = true}
```
onur-ozkan, line 34: As far as I can see, this crate is just a temporary hack and I am unsure that it will be well-maintained in the future (especially since it's very unknown crate). Instead, we could simply downgrade the `bitflags` version which might be a cleaner solution, but I am opposed to that as well. If we change the major version, it should be totally acceptable to introduce breaking changes in the serialization/deserialization layouts. 

If `bitflags` has changed their ser/deser layout for good, the sooner we adapt to it the better.

pv42:
If the bit flags 2.x behaviour should be used from now on this can be closed and I' open a new PR for the tests later.

PR #360, opened 2025-08-15
parser optimizations

onur-ozkan:
`MavXmlFilter::filter` should be about twice as fast compared to the previous version, and invoking `to_pascal_case` should also be quite faster compared to the previous version as its inner logic is much cheaper and is now forced to be inlined (since it is frequently called in loops).

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -1651,18 +1651,25 @@ impl MavXmlFilter {
     }
 }
 
+#[inline(always)]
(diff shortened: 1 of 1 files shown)


onur-ozkan:
> The code looks fine, but since the goal is speed optimization, did you measure it?

No, I would have shared it otherwise. It is so obvious I didn't think of measuring it.

joaoantoniocardoso (approved):
I see. I've just quickly measured here w/ criterion, it's around 5x faster =)

Issue #364, opened 2025-08-18
#363 breaks CLI usage of `mavlink-bindgen`


pv42:
Issue for tracking this comment since the PR is already merged
See https://github.com/mavlink/rust-mavlink/pull/363#issuecomment-3197332065

onur-ozkan:
> > This still covers only official dialects, when trying to generate custom dialects this would not work.
> 
> Right, even with checking cli feature this woudln't work for people who use bindgen as a library. We should handle this differently then.

Just proposed a solution to fix this: https://github.com/mavlink/rust-mavlink/pull/368

PR #368, opened 2025-08-19
re-write entire codegen flow

onur-ozkan:
Refactors codegen logic to handle both file lists and directories
as MAVLink XML sources.

This simplifies the codegen flow, improves error handling and allows
to generate dialects selectively when used as a library.

Reverts https://github.com/mavlink/rust-mavlink/pull/363
Fixes https://github.com/mavlink/rust-mavlink/issues/364


onur-ozkan:
cc @pv42

PR #369, opened 2025-08-20
Validate that fields in messages are unique

azerupi:
Follow up on the discussion in #357 

This validates that fields in messages are unique before generating the bindings. Generating the bindings with duplicate fields would lead to structs containing duplicate fields, which would fail to compile.

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -758,6 +758,22 @@ impl MavMessage {
             }
         }
     }
+
+    /// Ensures that a message does not contain duplicate field names.
+    ///
+    /// Duplicate field names would generate invalid Rust structs.
+    fn validate_unique_fields(&self) {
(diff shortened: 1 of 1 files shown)


mavlink-bindgen/src/parser.rs
```diff
@@ -758,6 +758,22 @@ impl MavMessage {
...
+    fn validate_unique_fields(&self) {
+        let mut seen: HashSet<&str> = HashSet::new();
+        for f in &self.fields {
+            let name: &str = &f.name;
+            if !seen.insert(name) {
+                panic!(
+                    "Duplicate field '{}' found in message '{}' while generating bindings",
+                    name, self.name
+                );
+            }
```
onur-ozkan: Nit:

You can do `assert!(!seen.insert(name)), "Duplicate field '{name}' found..."`.

azerupi: Done :+1:

Issue #370, opened 2025-08-20
Char[] / String handling is a bit broken?


azerupi:
On the project I'm working on we are parsing MAVLink messages from PX4 and QGroundControl, intercepting some and forwarding the rest. While doing that I've noticed that some forwarded messages containing `char[]` fields are not byte equivalent before and after decode and re-encoding.

The MAVLink spec doesn't specify anything about NULL termination and each message seems to have its own specific requirements around it. Some examples:

- [**AUTH_KEY (7)**](https://mavlink.io/en/messages/common.html#AUTH_KEY): `key: char[32]` nothing specified. No null-termination as the key would always be 32 chars long.
- [**PARAM_VALUE (22)**](https://mavlink.io/en/messages/common.html#PARAM_VALUE): `param_id: char[16]` terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars
- [**ADSB_VEHICLE (246)**](https://mavlink.io/en/messages/common.html#ADSB_VEHICLE): `callsign: char[9]` 8+null
- [**STATUSTEXT (253)**](https://mavlink.io/en/messages/common.html#STATUSTEXT): `text: char[50]` Status text message, without null termination character

It is not clear to me if ALL strings that are smaller than the `char[]` field length should be null-terminated or not?

In any case rust-mavlink does not handle this correctly. If we look at the `PARAM_VALUE_DATA` struct that is generated:

```rust
pub struct PARAM_VALUE_DATA {
    // omitted some fields...
    pub param_id: arrayvec::ArrayString<16>,
    // omitted some fields...
}
```

In the deserialization code we can see that if a null byte is encountered it will be truncated from the string.
```rust
for _ in 0..16usize {
    let next_char = buf.get_u8();
    if next_char == 0 {
        break;
    }
    __struct.param_id.push(next_char as char);
}
```

In the serialization however we are only adding values for as many bytes are in the slice. First this is bad because it means if the array/string does not contain exactly the amount of bytes that the field is expecting we are shifting everything left. Second when re-encoding the message we received from the deserialization step we've already lost the null-byte if it was there and thus we are always triggering the issue.

```rust
let slice = self.param_id.as_bytes();
for val in slice {
    __tmp.put_u8(*val);
}
```

Here is an example of the `PARAM_VALUE` message before deserialization and after deserialization + re-serialization in wireshark. You can see the null byte is missing and the length is one shorter.

[...]

PR #371, opened 2025-08-20
Fix ser/deser of strings

azerupi:
Attempt at fixing #370 

This is an attempt at fixing the issues I have observed with the string serialization and deserialization. This seems to be fixing the issues I had but let's see if it would work in a more general case. I've done only a quick test, I'm opening this PR for visibility and discussion but there might be more work to be done.

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -1026,7 +1026,7 @@ impl MavType {
                 quote! {
                     let slice = #val.as_bytes();
                     let remaining_capacity = #val.remaining_capacity();
-                    for val in slice.iter().chain(std::iter::repeat(&0u8).take(remaining_capacity)) {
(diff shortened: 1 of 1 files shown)


mavlink-bindgen/src/parser.rs
```diff
@@ -1026,7 +1025,8 @@ impl MavType {
                 let w = Char.rust_writer(&quote!(*val), buf);
                 quote! {
                     let slice = #val.as_bytes();
-                    for val in slice {
+                    let remaining_capacity = #val.remaining_capacity();
+                    for val in slice.iter().chain(std::iter::repeat(&0u8).take(remaining_capacity)) {
```
joaoantoniocardoso: We can't use `std::iter::repeat` here, as we must be compliant with no-std builds.

azerupi: Indeed, I noticed the CI failure yesterday but didn't have the opportunity to fix until now. I changed it to use core, it should be good now.

PR #372, opened 2025-08-21
Implemented recv_raw for async connection

joaofrf:
Added an async recv_raw method to AsyncMavConnection and implemented it for AsyncFileConnection.
This provides a way to receive MAVLinkMessageRaw without fully parsing them into typed messages.

The motivation is that in some scenarios parsing is not necessary — for example:

- Performance-critical applications where avoiding the extra parsing step reduces overhead.
- Cases where consumers only need to log, forward, or inspect the raw frames.


joaofrf:
I agree, I thought about it when I finished it. I can take a look at it tomorrow or during the weekend

mavlink/tests/tcp_loopback_async_tests.rs
```diff
@@ -28,6 +29,8 @@ mod test_tcp_connections {
             #[cfg(feature = "signing")]
             server.setup_signing(Some(singing_cfg_server));
 
+            // mavlink::commo
+
```
onur-ozkan: Seems like a leftover.

mavlink/tests/tcp_loopback_tests.rs
```diff
@@ -69,4 +70,65 @@ mod test_tcp_connections {
 
         server_thread.join().unwrap();
     }
+
+    /// Test whether we can send a message via TCP and receive it OK as a raw message. This also test signing as a property of a MavConnection if the signing feature is enabled.
+    #[test]
+    pub fn test_tcp_loopback_recv_raw() {
```
onur-ozkan: No need to `pub` a test function.

mavlink/tests/tcp_loopback_async_tests.rs
```diff
@@ -70,4 +73,69 @@ mod test_tcp_connections {
 
         server_thread.await.unwrap();
     }
+
+    /// Test whether we can send a message via TCP and receive it OK using async_connect recv_raw.
+    /// This also test signing as a property of a MavConnection if the signing feature is enabled.
+    #[tokio::test]
+    pub async fn test_tcp_loopback_recv_raw() {
+        const RECEIVE_CHECK_COUNT: i32 = 5;
```
onur-ozkan, line 79: Same, no need to `pub` a test function.

mavlink/tests/udp_loopback_async_tests.rs
```diff
@@ -46,4 +47,48 @@ mod test_udp_connections {
         }
         assert_eq!(recv_count, RECEIVE_CHECK_COUNT);
     }
+
+    /// Test whether we can send a message via UDP and receive it OK using async_connect recv_raw
+    #[tokio::test]
+    pub async fn test_udp_loopback_recv_raw() {
```
onur-ozkan: Same.

mavlink/tests/udp_loopback_tests.rs
```diff
@@ -44,4 +46,46 @@ mod test_udp_connections {
         }
         assert_eq!(recv_count, RECEIVE_CHECK_COUNT);
     }
+
+    /// Test whether we can send a message via UDP and receive it OK using recv_raw
+    #[test]
+    pub fn test_udp_loopback_recv_raw() {
```
onur-ozkan: Same.

mavlink-core/src/connection/file.rs
```diff
@@ -67,6 +67,35 @@ impl<M: Message> MavConnection<M> for FileConnection {
         }
     }
 
+    fn recv_raw(&self) -> Result<MAVLinkMessageRaw, crate::error::MessageReadError> {
+        // TODO: fix that unwrap
+        // not simple b/c PoisonError is not simple
```
onur-ozkan: Please remove this TODO along with the one in the `recv` version.

onur-ozkan (changes requested):

mavlink/tests/tcp_loopback_async_tests.rs
joaofrf: Removed

mavlink/tests/udp_loopback_async_tests.rs
joaofrf: Removed pub

mavlink/tests/tcp_loopback_tests.rs
joaofrf: Removed pub

mavlink/tests/tcp_loopback_async_tests.rs
joaofrf: Removed pub

mavlink-core/src/connection/file.rs
joaofrf: I've removed the TODO message from both methods

mavlink/tests/udp_loopback_tests.rs
joaofrf: Removed pub

mavlink/tests/tcp_loopback_async_tests.rs
joaofrf: not sure why github it not marking this as outdated "https://github.com/mavlink/rust-mavlink/pull/372/commits/df03798fa5b6650326d6c7bf08d09219fcc6f850" it is changed here - probably because line number changed?

mavlink-core/src/connection/file.rs
```diff
@@ -67,6 +66,34 @@ impl<M: Message> MavConnection<M> for FileConnection {
         }
     }
 
+    fn recv_raw(&self) -> Result<MAVLinkMessageRaw, crate::error::MessageReadError> {
+        // not simple b/c PoisonError is not simple
+        let mut file = self.file.lock().unwrap();
```
pv42, line 69: This comment seems to the second part of the removed todo, I don't think just removing the first line solves the issue. 
A proper solution would be to either justify why the unwrap() can not panic e.g. with an expect(..) or document that it might do so.
Alternativly at least open an issue so that it can be tracked that this need to be resolved.

onur-ozkan: It should be fine to panic on poisoned mutex. We already unwrap mutexes on other parts of the codebase.

But yes, this line should be removed as well.

joaofrf: I'll remove that comment too, it was copied from recv method so I thought someone might be looking into, or at least in the future.

PR #374, opened 2025-08-22
Test more non std

joaoantoniocardoso:
More tests available for `--no-default-features --features=embedded,common`


mavlink/tests/agnostic_decode_test.rs
```diff
@@ -75,7 +75,7 @@ mod test_agnostic_encode_decode {
     }
 }
 
-#[cfg(all(feature = "std", feature = "tokio-1", feature = "common"))]
+#[cfg(all(feature = "tokio-1", feature = "common"))]
```
onur-ozkan: Tokio without std would fail if we are not enabling std automatically on tokio feature.

joaoantoniocardoso: That's a good point, I'll undo this one

PR #377, opened 2025-08-22
Fix arbitrary feature

joaoantoniocardoso:
Depends on #376 (merged)

This patch makes it build again for `--features=arbitrary`, and it also adds the feature to the tests so we don't break it again until we actually start using it in tests (implemented in the next PR).


onur-ozkan (approved):
Seems valid change, thanks

Issue #291, opened 2024-11-16
Enumeration of all variants of MavMessage as list of Values (u32 or &str)


federico123579:
Currently `mavlink-bindgen` applied to a dialect generates the struct `MavMessage` with all variants available, though listing all these variants and collect them as a `Vec<&'static str>` or `Vec<u32>` seems still not possible. I need a way to iterate through all variants without using a match (in my case a GUI should offer a list of message names to choose from). Even a simple [`#derive(EnumIter)`](https://docs.rs/strum/latest/strum/derive.EnumIter.html) from [**Strum**](https://docs.rs/strum/latest/strum) would be enough for me.

PR #378, opened 2025-08-22
Test all messages for all dialects: default and random

joaoantoniocardoso:
Depends on #377


mavlink-core/src/lib.rs
```diff
@@ -144,6 +144,11 @@ where
     fn extra_crc(id: u32) -> u8;
 }
 
+pub trait MavlinkDialect {
+    /// All messages IDs
+    const MESSAGE_IDS: &'static [u32];
+}
+
```
onur-ozkan: This seems like isn't useful for users. Maybe define it for tests only?

joaoantoniocardoso: Maybe put this behind some other feature? Once I was trying to implement runtime dialect recognition (like what mavp2p does), and this was something I was missing... And although not precisely what was asked, check #291.

onur-ozkan: Oh, so it is useful in some cases. Wouldn't it be better to provide this information under `impl MavMessage` instead? A function like `const fn all_ids(&self) -> &'static [u32]` would be more accessible to users since they could see it directly on `MavMessage` without needing to know about/import the `*Dialect` types. It should be less complexity to maintain as well since we don't have to introduce new trait and types.

joaoantoniocardoso: Good idea, I'll try that, thanks!

joaoantoniocardoso: Done, but it looks like GitHub Actions is unstable now and failing to test it.

joaoantoniocardoso: ok, now it's back working again

mavlink-core/src/lib.rs
```diff
@@ -103,6 +103,18 @@ pub use connection::file::config::FileConfig;
...
+///
+/// Each message sets `MavMessage` enum implements this trait. The [`Dialect`] trait is used to
+/// represent dialects in an abstract way.
+pub trait Dialect
+where
+    Self: Sized,
+{
+    /// All dialect's messages IDs
+    const MESSAGE_IDS: &'static [u32];
+}
```
onur-ozkan: I guess I wasn't clear. What I meant was writing a `const fn` for each `MavMessage`, like:

```rs
impl MavMessage {
    const fn all_ids(&self) -> &'static [u32] {
        ...
    }
}
```

without introducing any additional types or variables.

PR #379, opened 2025-08-22
Change return type on message functions to Option

pv42:
`message_id_from_name()`, `default_message_from_id()` and `random_message_from_id()` only have one possible error: the message id or name does not exist, therefore `Option<_>` is a more appropriate type.


joaoantoniocardoso (approved):

onur-ozkan (approved):
Makes sense, thanks!

PR #380, opened 2025-08-22
Add tests to validate decode encode roundtrip

azerupi:
This adds only the regression tests that were added in #371 to validate the decoding and re-encoding a MAVLink message produces the same byte sequence. Whatever solution we land on, these tests are relevant to avoid future regressions.

The tests use two messages captured from a real communication between PX4 and QGC that contain a `char[]` field that is not the latest field (to avoid truncation of null bytes) and where strings are shorter than the `char[]` array.

### DIFF
--- a/mavlink/tests/v2_encode_decode_tests.rs
+++ b/mavlink/tests/v2_encode_decode_tests.rs
@@ -4,6 +4,7 @@ mod test_shared;
 mod test_v2_encode_decode {
     use crate::test_shared::HEARTBEAT_V2;
     use mavlink_core::peek_reader::PeekReader;
+    use mavlink_core::Message;
 
     #[test]
     pub fn test_read_v2_heartbeat() {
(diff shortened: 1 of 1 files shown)


mavlink/tests/v2_encode_decode_tests.rs
```diff
@@ -166,4 +167,53 @@ mod test_v2_encode_decode {
...
+        0x00, 0x00, 0x00, 0x00, 0x05, 0x87, 0x87,
+    ];
+
+    #[test]
+    pub fn test_decode_encode_v2_frame_parameter_value_bat1_r_internal() {
+        let mut r = PeekReader::new(PARAMETER_VALUE_BAT1_R_INTERNAL);
+        let (header, msg) =
+            mavlink::read_v2_msg::<mavlink::common::MavMessage, _>(&mut r).expect("decode");
+
+        let mut out = Vec::new();
```
joaoantoniocardoso: Could you make it also compatible with embedded? (see #374)

azerupi: Is this enough or are there other things to change?

joaoantoniocardoso: That's it, ty!

joaoantoniocardoso (approved):

onur-ozkan (approved):

PR #381, opened 2025-08-22
Add assert to check payload length during codegen

pv42:
Both MAVLink versions only support up to 255 payload bytes.
Currently the codegenerator will happily generate messages with any `usize` length. Such messages will cause panics when attempting to de-/serialize them. 

This PR adds an assert to prevent such cases.

Note: It seems to be theoretically possbile to have a message definition over 255 bytes fit into a 255 payload with truncated zeros, but I cant see a usecase arise from this since everything past byte 255 would always need to be 0.

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -702,9 +702,9 @@ impl MavMessage {
         let id = self.id;
         let name = self.name.clone();
         let extra_crc = extra_crc(self);
-        let (name_types, msg_encoded_len) = self.emit_name_types();
+        let (name_types, payload_encoded_len) = self.emit_name_types();
         assert!(
-            msg_encoded_len <= 255,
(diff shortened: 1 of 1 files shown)


mavlink-bindgen/src/parser.rs
```diff
@@ -703,6 +703,10 @@ impl MavMessage {
         let name = self.name.clone();
         let extra_crc = extra_crc(self);
         let (name_types, msg_encoded_len) = self.emit_name_types();
+        assert!(
+            msg_encoded_len <= 255,
```
joaoantoniocardoso: Is encoded_len the payload and not the entire message? 🤔 that's confusing

pv42: I update the name to better express what it does

onur-ozkan (approved):

PR #382, opened 2025-08-25
Expose inner buffer

### LINKED ISSUE #203: Allow access to the internal buffers of raw messages
Currently, the `MAVLinkV1MessageRaw` and the `MAVLinkV2MessageRaw` types prevent mutable access to the internal buffers they use to store the raw bytes coming from any type of connection, which is problematic for libraries that built on top of this crate. 

What is actually being problematic is that this crate *enforces* you to use the *synchronous* reads (e.g `read_v2_raw_message`) in order to *fill* the raw message buffer *using* a reader, and there is no other way.

Building functionality over these raw packets is *exclusive* to the `rust-mavlink` library at the moment, preventing other libraries to interact with `rust-mavlink` in a flexible way. For example, to be able to implement a new `read()` function (e.g an async) for MAVLink messages from *another* library, access to the internal buffer of a raw message is required, which is not possible at the moment.

I highly suggest that these types *somehow* allow mutable access to the internal buffers.

joaoantoniocardoso:
This allows external parsers to be used with this library.

Close #203

### DIFF
--- a/mavlink-core/src/lib.rs
+++ b/mavlink-core/src/lib.rs
@@ -427,6 +427,31 @@ impl MAVLinkV1MessageRaw {
         Self([0; 1 + Self::HEADER_SIZE + 255 + 2])
     }
 
+    /// Create a new raw MAVLink 1 message from a given buffer.
+    ///
+    /// Note: This method does not guarantee that the constructed MAVLink message is valid.
+    pub const fn from_bytes_unparsed(bytes: [u8; 1 + Self::HEADER_SIZE + 255 + 2]) -> Self {
+        Self(bytes)
+    }
+
+    /// Read access to its internal buffer.
+    #[inline]
+    pub fn as_slice(&self) -> &[u8] {
+        &self.0
+    }
+
+    /// Mutable reference to its internal buffer.
+    #[inline]
+    pub fn as_mut_slice(&mut self) -> &mut [u8] {
+        &mut self.0
+    }
+
+    /// Deconstruct the MAVLink message into its owned internal buffer.
+    #[inline]
+    pub fn into_inner(self) -> [u8; 1 + Self::HEADER_SIZE + 255 + 2] {
+        self.0
+    }
+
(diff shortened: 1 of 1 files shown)


mavlink-core/src/lib.rs
```diff
@@ -775,11 +800,38 @@ impl MAVLinkV2MessageRaw {
...
+    /// Create a new raw MAVLink 2 message filled with zeros.
     pub const fn new() -> Self {
         Self([0; 1 + Self::HEADER_SIZE + 255 + 2 + Self::SIGNATURE_SIZE])
     }
 
+    /// Create a new raw MAVLink 1 message from a given buffer.
+    ///
+    /// Note: This method does not guarantee that the constructed MAVLink message is valid.
+    pub fn from_bytes_unparsed(
+        bytes: [u8; 1 + Self::HEADER_SIZE + 255 + 2 + Self::SIGNATURE_SIZE],
```
onur-ozkan: This can be a const function to allow it be constructed in a constant context (e.g., const/global variables, const functions) if needed.

mavlink-core/src/lib.rs
```diff
@@ -422,11 +422,36 @@ impl Default for MAVLinkV1MessageRaw {
...
     pub const fn new() -> Self {
         Self([0; 1 + Self::HEADER_SIZE + 255 + 2])
     }
 
+    /// Create a new raw MAVLink 1 message from a given buffer.
+    ///
+    /// Note: This method does not guarantee that the constructed MAVLink message is valid.
+    pub fn from_bytes_unparsed(bytes: [u8; 1 + Self::HEADER_SIZE + 255 + 2]) -> Self {
+        Self(bytes)
+    }
```
onur-ozkan: Same here.

mavlink-core/src/lib.rs
joaoantoniocardoso: Agreed!

PR #384, opened 2025-08-26
Fix `cargo doc` for stable rust

### LINKED ISSUE #361: `cargo doc` failing on stable Rust
When trying to run `cargo doc` command it fails as the `#![feature]` is not available on stable Rust versions. Here is the error message that pop ups when running the command

```
PS C:\Users\user\Documents\code\test_mavlink> cargo doc
 Documenting bitflags v2.9.1
 Documenting mavlink-core v0.15.0
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> C:\Users\user\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\mavlink-core-0.15.0\src\lib.rs:23:50
   |
23 | #![cfg_attr(all(any(docsrs, doc), not(doctest)), feature(doc_auto_cfg))]
   |                                                  ^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0554`.
```

We are currently running Rust 1.87 and we would like to keep using the stable versions of Rust and not the nightly build. Is it possible to fix it as we also need the documentation tool in our project

pv42:
Fixes #361.

- only enable `doc_auto_cfg` feature for `docsrs` and no longer also for `doc`
- configure doc in the CI to use `--cfg=docsrs` to create documentation with nightly features as before (using the `doc_auto_cfg` feature)

Drawback is that simply using `cargo +nightly doc` will no longer enable the feature

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -128,7 +128,7 @@ jobs:
     - uses: actions/checkout@v5
     - uses: dtolnay/rust-toolchain@nightly
     - name: Build docs
-      run: cargo doc --features "default all-dialects emit-description emit-extensions format-generated-code tokio-1 signing" 
+      run: RUSTDOCFLAGS=--cfg=docsrs cargo doc --features "default all-dialects emit-description emit-extensions format-generated-code tokio-1 signing" 
     - name: Deploy
       uses: peaceiris/actions-gh-pages@v3
       if: ${{ github.ref == 'refs/heads/master' }}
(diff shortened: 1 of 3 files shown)


onur-ozkan:
> Drawback is that simply using cargo +nightly doc will no longer enable the feature

We can detect if it's being by nightly compiler and then enable the feature.

pv42:
> We can detect if it's being compiled by nightly the compiler and then enable the feature.

I considered that but discarded the idea since the main way to do so would with a build script and while `mavlink` already has one adding one to the other two crates seemed excessive for just enabling this relatively minor enhancement to the documentation.

But if you think that would be better or have another idea I am happy implement that.

PR #387, opened 2025-08-27
Make generated order deterministic and add snapshot testing

azerupi:
This changes the hashmaps to indexmaps that preserve order of insertion when iterating over them. This makes the generated code order deterministic and consistent with the definition file. Before this, multiple runs would provide different orders for the generated structs and enums.

This PR also adds snapshot testing using [cargo insta](https://insta.rs/docs/) to ensure that the generated code remains consistent across changes or makes it very obvious when it does change. This should make PR reviews slightly easier because you can see changes to generated code in the PR diffs.

I've intentionally kept amount of messages being snapshot tested minimal. They can be added on a case by case basis after this PR to exercise different parts of the generator.

### DIFF
--- a/mavlink-bindgen/Cargo.toml
+++ b/mavlink-bindgen/Cargo.toml
@@ -23,6 +23,7 @@ clap = { version = "~4.3.24", optional = true, default-features =false, features
 thiserror = { version = "2.0.12", default-features = false }
 arbitrary = { version = "1.4", optional = true, features = ["derive"] }
 rand = { version = "0.9", optional = true, default-features = false, features = ["std", "std_rng"] }
+indexmap = "2.11.0"
 
 # Required to support MSRV 1.65.0
 clap_lex = { version = "=0.7.5", optional=true }
@@ -38,3 +39,8 @@ cli = ["dep:clap", "dep:clap_lex", "dep:clap_builder", "dep:anstyle", "dep:ansty
 arbitrary = ["dep:arbitrary", "dep:rand"]
 emit-extensions = []
 emit-description = ["dep:regex"]
+serde = ["dep:serde", "indexmap/serde"]
+
+[dev-dependencies]
+insta = { version = "1.39.0", features = ["glob"] }
+tempfile = "3.10.1"
--- a/mavlink-bindgen/Cargo.toml
+++ b/mavlink-bindgen/Cargo.toml
@@ -23,7 +23,6 @@ clap = { version = "~4.3.24", optional = true, default-features =false, features
 thiserror = { version = "2.0.12", default-features = false }
 arbitrary = { version = "1.4", optional = true, features = ["derive"] }
 rand = { version = "0.9", optional = true, default-features = false, features = ["std", "std_rng"] }
-indexmap = "2.11.0"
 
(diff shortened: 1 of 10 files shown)


mavlink-bindgen/src/lib.rs
```diff
@@ -98,7 +98,7 @@ pub fn generate<P1: AsRef<Path>, P2: AsRef<Path>>(
                     continue;
                 }
 
-                bindings.push(generate_single_file(&entry.path(), &destination_dir)?);
+                bindings.push(generate_single_file(entry.path(), destination_dir)?);
```
azerupi: This was a clippy lint warning

mavlink-bindgen/Cargo.toml
azerupi, line 1: Sorry for the noise here, only the dev dependencies have been added. The rest was formatted by `cargo add`. I left it because next time people use `cargo add` it will stay consistent. But if you want me to revert the formatting changes I can do that.

azerupi:
I'm a bit confused by the CI failure. Running the same command locally doesn't provide the same result and grepping the source for `all_ids` doesn't seem to provide a match. Where is this code generated and under which conditions?

pv42:
Did you rebase your local on master ? Last time we had weird CI failures the reason was GitHub doing weird merges.

mavlink-bindgen/Cargo.toml
joaoantoniocardoso: Could you add the dependencies (the indexmap and dev-dependencies) without reformatting the file?

joaoantoniocardoso (commented):
Hi, I like the idea, but I'm not familiar with snap. Could you add a doc explaining how you generate them?

azerupi:
@onur-ozkan wouldn't a `BTreeMap` give you alphabetical order based on message names?
We can do it, it would solve the deterministic part. But in my opinion there is also value in having them ordered in order of appearance in the definition file. It makes it a little easier when browsing generated docs for example or following the types and landing in the generated files through rust-analyzer.

onur-ozkan:
> @onur-ozkan wouldn't a `BTreeMap` give you alphabetical order based on message names? We can do it, it would solve the deterministic part. But in my opinion there is also value in having them ordered in order of appearance in the definition file. It makes it a little easier when browsing generated docs for example or following the types and landing in the generated files through rust-analyzer.

Should we really care about syncing the order of definition files with the generated Rust code? Honestly having a consistent alphabetical order automatically would be pretty nice (e.g., if the order in the definition file gets messy).

Also since the definition files are XML, I always feel a bit lost when looking at them. I can't really say "this message was at position X in the definition file but at position Y in the Rust code", I am not sure if anyone would compare/care them.

Adding a new dependency is a big deal and I don't think this is worth for doing it.

azerupi:
I usually look at the mavlink docs online. Having the same order in the code is convenient for me. But I do get your point and I agree that it might not be worth it.

I'll make the changes.

azerupi:
@onur-ozkan the changes have been made. Anything else? :slightly_smiling_face:

joaoantoniocardoso (approved):

PR #388, opened 2025-08-27
Add wrapper for char arrays to access str conveniently

azerupi:
Follow-up on #370 

This adds a wrapper around char arrays that allows to access them either as slices or as strings. When accessing the string representation, it will automatically convert the char array to a string slice up to the first null byte.

A part I have some slight doubts about is the serde derives on the `CharArray<N>` struct. If someone could validate that this is correct that would be great.

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -1127,7 +1127,7 @@ impl MavType {
             "Double" => Some(Double),
             "double" => Some(Double),
             _ if s.starts_with("char[") => {
-                let start = s.find('[')?;
+                let start = 4;
                 let size = s[start + 1..(s.len() - 1)].parse::<usize>().ok()?;
                 Some(CharArray(size))
             }


onur-ozkan:
I will try to allocate sometime at Sunday to review this.

mavlink-core/src/types.rs
```diff
@@ -0,0 +1,147 @@
...
+/// use mavlink_core::types::CharArray;
+///
+/// let data = [0x48, 0x45, 0x4c, 0x4c, 0x4f, 0x00, 0x57, 0x4f, 0x52, 0x4c, 0x44, 0x00, 0x00, 0x00];
+/// let ca = CharArray::new(data);
+/// assert_eq!(ca.to_str(), "HELLO");
+/// ```
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[cfg_attr(feature = "serde", derive(Serialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+pub struct CharArray<const N: usize> {
```
onur-ozkan:
```suggestion
#[cfg_attr(feature = "serde", derive(Serialize))]
#[cfg_attr(feature = "serde", serde(transparent))]
#[derive(Debug, PartialEq, Clone, Copy)]
pub struct CharArray<const N: usize> {
```

mavlink-core/src/types.rs
```diff
@@ -0,0 +1,147 @@
...
+            data,
+            str_len: first_null,
+        }
+    }
+
+    /// Get the string representation of the char array.
+    /// Returns the string stopping at the first null byte and if the string is not valid utf8
+    /// the returned string will be empty.
+    pub fn to_str(&self) -> &str {
+        core::str::from_utf8(&self.data[..self.str_len]).unwrap_or("")
```
onur-ozkan: Why is that we return empty string if the value is invalid? I think the caller should decide what to do on invalid UTF8 values. We can simply return the inner `Result`.

onur-ozkan:
By any chance, would you be able to finalize this PR anytime soon or should we take it over? It would be great to have it included in the next release (which will be in a week or so).

mavlink-bindgen/src/parser.rs
```diff
@@ -715,6 +715,10 @@ impl MavMessage {
                         #[cfg_attr(feature = "serde", serde(with = "serde_arrays"))]
                         #[cfg_attr(feature = "ts", ts(type = "Array<number>"))]
                     )
+                } else if matches!(field.mavtype, MavType::CharArray(_)) {
+                    quote!(
+                        #[cfg_attr(feature = "ts", ts(type = "string"))]
```
patrickelectric, line 720: This appears to be wrong since the serde serialization does not change in this PR.

azerupi: So it should remain `Array<number>`?
```rust
#[cfg_attr(feature = "ts", ts(type = "Array<number>"))]
```

patrickelectric: Yes, since this is what serde will output.

azerupi: Done, let me know if there is anything else that needs to change for the ts bindings.

mavlink-bindgen/src/parser.rs
```diff
@@ -1133,16 +1126,18 @@ impl MavType {
...
-                if s.ends_with(']') {
-                    let start = s.find('[')?;
-                    let size = s[start + 1..(s.len() - 1)].parse::<usize>().ok()?;
-                    let mtype = Self::parse_type(&s[0..start])?;
-                    Some(Array(Box::new(mtype), size))
-                } else {
-                    None
-                }
+            _ if s.starts_with("char[") => {
+                let start = s.find('[')?;
```
joaoantoniocardoso: since it starts with `"char["`, the `[` position is already known (4)

azerupi: Good point, changed.

mavlink-bindgen/src/parser.rs
patrickelectric: Just pointing that since this is using #415 as base, the type should be string again.

azerupi: Yes, this has been changed in the latest commits 🙂

PR #389, opened 2025-08-28
Use deprecation tags in message definition for code generation

pv42:
In the MAVLink dialect files mesages, enums and enum fields can be marked as deprecated with 
`<deprecated since="YYYY-MM" replaced_by="...">optional description text</deprecated>` ([doc](https://mavlink.io/en/guide/xml_schema.html#deprecated)).

Currently if such a tag is encountered only a `eprintln!("TODO: deprecated {text}");` was used (is most cases at least).

This PR generates `#[deprecated = "..."]` attributes on all enums, enum values, message `*_DATA` structs and `MavMessage` enum entries that have such a tag.

I have also added a snapshot test that test a (shortened) example from common.xml for all 3 possible deprecated tag positions.

For the generated files to function as before without generating a lot of deprecation warnings they are now `#![allow(deprecated)]`. 

The documentation also says that some code-generators can optionally disable emitting deprecated messages enums and enum entries. This could also be done at a later point, for example with a by default enabled `emit-deprecated` feature.

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -112,6 +112,7 @@ impl MavProfile {
     //    }
 
     /// Simple header comment
+    #[inline(always)]
     fn emit_comments(&self, dialect_name: &str) -> TokenStream {
         let message = format!("MAVLink {dialect_name} dialect.");
         quote!(
@@ -122,15 +123,18 @@ impl MavProfile {
     }
 
     /// Emit rust messages
+    #[inline(always)]
     fn emit_msgs(&self) -> Vec<TokenStream> {
         self.messages.values().map(|d| d.emit_rust()).collect()
     }
 
     /// Emit rust enums
+    #[inline(always)]
     fn emit_enums(&self) -> Vec<TokenStream> {
         self.enums.values().map(|d| d.emit_rust()).collect()
     }
 
+    #[inline(always)]
     fn emit_deprecations(&self) -> Vec<TokenStream> {
         self.messages
             .values()
@@ -144,6 +148,7 @@ impl MavProfile {
     }
 
     /// Get list of original message names
+    #[inline(always)]
     fn emit_enum_names(&self) -> Vec<TokenStream> {
         self.messages
             .values()
@@ -155,6 +160,7 @@ impl MavProfile {
     }
 
     /// Emit message names with "_DATA" at the end
+    #[inline(always)]
     fn emit_struct_names(&self) -> Vec<TokenStream> {
         self.messages
             .values()
@@ -236,6 +242,7 @@ impl MavProfile {
         }
     }
 
+    #[inline(always)]
     fn emit_mav_message(
         &self,
         deprecations: &[TokenStream],
@@ -253,6 +260,7 @@ impl MavProfile {
         }
     }
 
+    #[inline(always)]
     fn emit_mav_message_all_ids(&self) -> TokenStream {
         let mut message_ids = self.messages.values().map(|m| m.id).collect::<Vec<u32>>();
         message_ids.sort();
@@ -264,6 +272,7 @@ impl MavProfile {
         )
     }
 
+    #[inline(always)]
     fn emit_mav_message_parse(
         &self,
         enums: &[TokenStream],
@@ -283,6 +292,7 @@ impl MavProfile {
         }
     }
 
+    #[inline(always)]
     fn emit_mav_message_crc(&self, id_width: &Ident, structs: &[TokenStream]) -> TokenStream {
         quote! {
             fn extra_crc(id: #id_width) -> u8 {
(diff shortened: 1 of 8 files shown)


mavlink/tests/helper_tests.rs
```diff
@@ -9,6 +9,7 @@ mod helper_tests {
         let id = id.unwrap();
         assert!(id == 4, "Invalid id for message name: PING");
         let message = MavMessage::default_message_from_id(id);
+        #[allow(deprecated)]
```
onur-ozkan: This should be better coverage here (and on others as well):

```suggestion
        #[expect(deprecated)]
```

mavlink-bindgen/src/parser.rs
```diff
@@ -1139,6 +1190,36 @@ impl MavType {
...
+}
+
+impl MavDeprecation {
+    pub fn emit_tokens(&self) -> TokenStream {
+        let since = &self.since;
+        let note = match &self.note {
+            Some(str) if str.is_empty() || str.ends_with(".") => str,
+            Some(str) => &format!("{str}."),
+            None => &String::new(),
+        };
```
joaoantoniocardoso:
```suggestion
        let note = match &self.note {
            Some(str) if str.is_empty() || str.ends_with(".") => str.clone(),
            Some(str) => format!("{str}."),
            None => String::new(),
        };
```

mavlink-bindgen/src/parser.rs
```diff
@@ -433,6 +453,12 @@ impl MavEnum {
                 let name = format_ident!("{}", enum_entry.name.clone());
                 let value;
 
+                let deprecation = enum_entry
+                    .deprecated
+                    .as_ref()
+                    .map(|d| d.emit_tokens())
+                    .unwrap_or_default();
```
onur-ozkan:
```suggestion
                let deprecation = self.emit_deprecation();
```

mavlink-bindgen/src/parser.rs
```diff
@@ -478,11 +506,20 @@ impl MavEnum {
         quote!(pub const DEFAULT: Self = Self::#default;)
     }
 
+    fn emit_deprecation(&self) -> TokenStream {
```
onur-ozkan, line 526:
```suggestion
    #[inline(always)]
    fn emit_deprecation(&self) -> TokenStream {
```

onur-ozkan (commented):
2 nits, LGTM otherwise.

Nice work btw!

mavlink-bindgen/src/parser.rs
pv42: self.emit_deprecation is the enum while enum_entry is a possible enum value

onur-ozkan: Ah, I see. I would get rid of them (emit_deprecation functions) at this point, to avoid confusion (feel free to leave if you think otherwise).

pv42: I added a `emit_deprecation` to `MavEnumEntry` so this is more clear

mavlink-bindgen/src/parser.rs
```diff
@@ -222,14 +236,19 @@ impl MavProfile {
         }
     }
 
-    fn emit_mav_message(&self, enums: &[TokenStream], structs: &[TokenStream]) -> TokenStream {
+    fn emit_mav_message(
```
onur-ozkan, line 246: All these small functions being called very frequently in quite massive loops, right? They should be inlined with "always" attribute.

pv42: Since these are private, small function with only one caller the compiler almost certainly inlines them already. 
I would only add more #[inline] if it is shown that this is not happening and that it improves performance.
These functions are called a lot but not enough that I would expect the performance impact to be over a few milliseconds. 
Regardless I would prefer not to add inlining to this PR since it is about the deprecation tags and not about performance improvements.

joaoantoniocardoso (approved):

mavlink-bindgen/src/parser.rs
onur-ozkan: > Since these are private, small function with only one caller the compiler almost certainly inlines them already.
> I would only add more #[inline] if it is shown that this is not happening and that it improves performance.

You never know that tbh and that's precisely the reason why we have `inline(always)` in the language. Without debugging extremely verbose compilation logs or the binary output, how would you know "when it's needed and when it's not" ? If something very small is being called very often from the same place, I don't know what would be a good reason to not inline it explicitly on the source code.

> Regardless I would prefer not to add inlining to this PR since it is about the deprecation tags and not about performance improvements.

I can see that this PR isn't about performance but it feels inconsistent to have that tag on some functions and not others. Besides, we shouldn't add things now, leave it and wish for some other people to optimize them later.

That's just my evaluation of this topic, feel free to not do it. @joaoantoniocardoso or @patrickelectric can also merge this.

joaoantoniocardoso: I'd just follow what's conveniently being practiced around that code: since all seem to be inlined for some previous thought reason, it shouldn't hurt to follow it.

If we suspect that inlining is not the best practice, we can evaluate that more broadly, considering all the other functions too (but not in this PR).

PR #390, opened 2025-09-02
Use version and dialect tags in code generation

pv42:
This PR resolves the last 2 regularly occurring `eprintln!("TODO: ...")` in the code generation. The `<version>` and `<dialect>` tags are properly handled. Both tags are optional.

- The dialect tag which is a unique dialect number is emitted as a `pub const DIALECT_NUMBER: `~~`Option<u8>`~~ `u8`
- If the dialect is not set the constant will ~~be `None`~~ not get generated
- The version tag which represent the minor MAVLink version is emitted as `pub const MINOR_MAVLINK_VERSION: `~~`Option<u8>`~~ `u8`
- If the version is not set it is attempted to use the first valid one from an `<include>`, otherwise it is also ~~`None`~~ not get generated
- If either of those numbers can not be parsed as a `u8` code generation panics.  

The minor version is supposed to be used in the "uint8_t_mavlink_version" magic type. This field is supposed to not be user writable. The current implementation treats it the same as `unit8_t`.
In this PR currently only the `impl Default` for messages with this type (usually only HEARTBEAT) is changed, but alternatively the `ser` function could be also adjusted.

### Tests
- add version and dialect to heartbeat test
- adjust other snapshoot test
- adjust test that uses default heartbeat

[MAVLink XML doc](https://mavlink.io/en/guide/xml_schema.html)

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -43,6 +43,8 @@ lazy_static! {
 pub struct MavProfile {
     pub messages: BTreeMap<String, MavMessage>,
     pub enums: BTreeMap<String, MavEnum>,
+    pub version: Option<u8>,
+    pub dialect: Option<u8>,
 }
 
 impl MavProfile {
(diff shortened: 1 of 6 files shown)


onur-ozkan:
Would it make more sense to omit `const` var entirely instead of setting it to `None` when it's not available? Is there a current use-case where `Option` would be useful?

pv42:
> Would it make more sense to omit `const` var entirely instead of setting it to `None` when it's not available? Is there a current use-case where `Option` would be useful?

I changed it so they only get generated if they are defined. This makes the generated code nicer but whether a `<version>` exists is passed around a bit internally for the default value it know if it can use the const.

onur-ozkan (approved):

PR #392, opened 2025-09-04
Hide `bytes` and `bytes_mut` in doc

pv42:
These are not intended to be a part of the public api, they are only `pub` so they can be used by the generated dialect code.

### DIFF
--- a/mavlink-core/src/lib.rs
+++ b/mavlink-core/src/lib.rs
@@ -43,7 +43,9 @@ use crate::{bytes::Bytes, error::ParserError};
 
 use crc_any::CRCu16;
 
+#[doc(hidden)]
 pub mod bytes;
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):

mavlink-core/src/lib.rs
```diff
@@ -43,7 +43,9 @@ use crate::{bytes::Bytes, error::ParserError};
 
 use crc_any::CRCu16;
 
+#[doc(hidden)]
 pub mod bytes;
+#[doc(hidden)]
 pub mod bytes_mut;
```
onur-ozkan, line 49: ~Is there a reason why they are `pub` then?~ I just saw the PR desc :)

PR #396, opened 2025-09-12
switch non-blocking mode in tcp&udp sockets

onur-ozkan:
`try_recv` wasn't really blocking before and this should do it.


onur-ozkan:
Do we have any concerns on merging this?

joaoantoniocardoso (approved):

Issue #386, opened 2025-08-27
documenting bindgen generated code


onur-ozkan:
Things we parse in `mavlink-bindgen/src/parser.rs` should be documented. Some of them will require parsing message descriptions from the message definition file, others can be documented manually.

PR #397, opened 2025-09-13
Emit more desc

onur-ozkan:
Resolves https://github.com/mavlink/rust-mavlink/issues/386 and removes the `emit-description` feature. That feature doesn't provide much value and most likely many users are unaware of its existence. It's not only useful for building docs, but also useful for coding as we can read the docs while inspecting types with rust-analyzer.

This PR improves documentation for `MavMessage` and formats the generated docs in a more readable way.


mavlink-bindgen/src/parser.rs
```diff
@@ -707,22 +720,20 @@ impl MavMessage {
...
-            } else {
-                &format!("{doc}.")
-            };
+            let doc = format!("{doc}{}", if doc.ends_with('.') { "" } else { "." });
             // create hyperlinks
-            let doc = URL_REGEX.replace_all(doc, "<$1>");
+            let doc = URL_REGEX.replace_all(&doc, "<$1>");
             ts.extend(quote!(#[doc = #doc]));
+            // Leave two blank lines before the message ID for readability.
+            ts.extend(quote!(#[doc = ""]));
```
pv42: What is the reason for double blank lines, doesn't a single one achive the same ?

onur-ozkan: Apparently it is, weird...

onur-ozkan:
Do we have any concerns on merging this?

patrickelectric (approved):
I see not problem with it. It may have a performance issue, but if it was tested and does not affect it much, the PR should be fine

PR #400, opened 2025-09-18
Properly handle message ids over 255 for MAVLink 1

pv42:
Currently writing a message with an id over 255 with MAVLink 1 will overflow the message id and subsequently send the data as normal. Due a CRC_EXTRA missmatch it is unlikely that the data will be received as valid but it is sent either way.

There does not seem to be a definition what should happen in such case in the standard and the reference implementations seem to handle it differently (python crashes, C seems to also overflow the ID) but either returning an error or panicing seem like valid. 

This PR makes it so that 
- the `write_v1_msg` functions returns a `MessageWriteError::MAVLink2Only`  and
- the lower level  `serialize_message` functions panic 

when attempting to write a message with an id over 255.

### DIFF
--- a/mavlink-core/src/lib.rs
+++ b/mavlink-core/src/lib.rs
@@ -39,7 +39,10 @@ use serde::{Deserialize, Serialize};
 pub mod peek_reader;
 use peek_reader::PeekReader;
 
-use crate::{bytes::Bytes, error::ParserError};
+use crate::{
+    bytes::Bytes,
+    error::{MessageWriteError, ParserError},
+};
 
 use crc_any::CRCu16;
 
@@ -611,6 +614,9 @@ impl MAVLinkV1MessageRaw {
         &self.0[..(1 + Self::HEADER_SIZE + payload_length + 2)]
     }
 
+    /// # Panics
+    ///
+    /// If the `msgid` parameter exceeds 255 and is therefore not supported for MAVLink 1
     fn serialize_stx_and_header_and_crc(
         &mut self,
         header: MavHeader,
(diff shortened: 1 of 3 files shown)


mavlink-core/src/lib.rs
```diff
@@ -626,7 +632,7 @@ impl MAVLinkV1MessageRaw {
             header.sequence,
             header.system_id,
             header.component_id,
-            msgid as u8,
+            msgid.try_into().unwrap(),
```
onur-ozkan, line 635: Why not taking u8 directly so we can do the validation with type system?

mavlink-core/src/lib.rs
```diff
@@ -1981,6 +2000,9 @@ pub async fn write_v1_msg_async<M: Message>(
     header: MavHeader,
     data: &M,
 ) -> Result<usize, error::MessageWriteError> {
+    if data.message_id() > u8::MAX.into() {
```
onur-ozkan, line 2003: Same here, this should be possible to handle with the type system directly.

mavlink-core/src/lib.rs
pv42: This function is used by only 2 functions.
`pub MAVLinkV1MessageRaw::serialize_message` and `pub MAVLinkV1MessageRaw::serialize_message_data` call it with `Message::message_id() -> u32` and `MessageData::ID: u32` which have to be u32 (or a least u24) since they also support MAVLink 2.

It would be possible to move the cast into those but that would not change anything fundamentally.

mavlink-core/src/lib.rs
pv42: Same here `data.message_id()` is also used by V2 which support 3 byte message ids

mavlink-core/src/lib.rs
onur-ozkan: Maybe we could re-design/refactor certain parts later to handle this. Thanks.

PR #404, opened 2025-09-22
mavlink: Add typescript generation

patrickelectric:


mavlink/src/lib.rs
```diff
@@ -55,6 +55,7 @@
 //! [tokio]: https://crates.io/crates/tokio
 //! [serde]: https://crates.io/crates/serde
 //! [arbitrary]: https://crates.io/crates/arbitrary
+//! [ts]: https://crates.io/crates/ts-rs
```
pv42, line 59: Also further up where all the features are listed

mavlink/Cargo.toml
```diff
@@ -38,6 +38,9 @@ serde_arrays = { version = "0.2.0", optional = true }
 arbitrary = { version = "1.4", optional = true, features = ["derive"] }
 rand = { version = "0.9", optional = true, default-features = false, features = ["std", "std_rng"] }
 
+# Used for typescript generation
+ts-rs = { version = "11.0.0", optional = true }
```
joaoantoniocardoso: I think this commentary is unnecessary, and this dependency can be kept together with the other ones.

mavlink/Cargo.toml
```diff
@@ -98,6 +101,9 @@ serde = ["bitflags/serde", "dep:serde", "dep:serde_arrays", "mavlink-core/serde"
 tokio-1 = ["mavlink-core/tokio-1"]
 arbitrary = ["dep:arbitrary", "dep:rand", "mavlink-bindgen/arbitrary", "mavlink-core/arbitrary", "bitflags/arbitrary"]
 
+# Used for typescript generation
+ts = ["dep:ts-rs"]
+
```
joaoantoniocardoso, line 103: maybe:

```suggestion
ts = ["dep:ts-rs"] # Used for typescript generation
```

and then keep it with the other ones?

joaoantoniocardoso (changes requested):
Nice!! I've been thinking of adding this integration myself for a while, so I have some thoughts here:

1. We should add a note somewhere saying that the bindings are generated by running `cargo test` with the `ts` feature enabled. If this is not what we want, we should remove all the `ts(export)` derive macros **and** either (a.) make a build.rs with a typescript bindings generation step conditionally enabled, or (b.) make an example or binary to generate the bindings, with simple instructions.
2. We should add `mavlink/bindings` (or whatever the destination would be) to our `.gitignore`.
3. I think this patch is missing the extensions (search for `emit-extensions`)
4. We must somehow ensure that the generated TypeScript is 1:1 compatible with our serde JSON feature. Maybe using an Example to show the integration? I'm open to ideas!

mavlink/Cargo.toml
patrickelectric: Done.

mavlink/Cargo.toml
patrickelectric: There is no inline comment in the entire file, I want to maintain it consistent.

mavlink/Cargo.toml
```diff
@@ -37,6 +37,7 @@ serde = { version = "1.0.115", optional = true, features = ["derive"] }
 serde_arrays = { version = "0.2.0", optional = true }
 arbitrary = { version = "1.4", optional = true, features = ["derive"] }
 rand = { version = "0.9", optional = true, default-features = false, features = ["std", "std_rng"] }
+ts-rs = { version = "11.0.0", optional = true }
```
joaoantoniocardoso: 11.0.1 seems to be the latest: [changelog](https://github.com/Aleph-Alpha/ts-rs/blob/main/CHANGELOG.md).

patrickelectric: Done.

patrickelectric:
ping @joaoantoniocardoso

mavlink/Cargo.toml
joaoantoniocardoso: maybe try again?

patrickelectric: done.

PR #407, opened 2025-09-23
Panic and error documentation

pv42:
This PR adds better error and documentation to `mavlink-core`

# Breaking

To have consistent `read_` function names I renamed the following group of functions:
`read_raw_versioned_msg[_async][_signed]` -> `read_versioned_raw_message[_async][_signed]`

# Changes

- `# Errors` and `# Panics` doc sections to all applicable public functions in `mavlink-core`
- Add unified doc for `read_` and `write_` functions and their errors in mavlink-core/src/lib.rs
- simplyfy doc link when applicable
- properly import commonly used items e.g. `crate::error::MessageReadError`

# Lints

There are 2 clippy lints that check for the existence of these sections but in my experience the one for panics is not very useful since it has many false positives and negatives. Therefore I did not enable them.

# Other

Only lock signing mutex when message actually needs to be signed, this is a minor performance improvement for the `signing` feature. This is so minor that I didn't want to put it in a separate PR.

### DIFF
--- a/mavlink-core/src/async_peek_reader.rs
+++ b/mavlink-core/src/async_peek_reader.rs
@@ -57,11 +57,17 @@ impl<R: AsyncReadExt + Unpin, const BUFFER_SIZE: usize> AsyncPeekReader<R, BUFFE
     /// If the internal buffer does not contain enough data, this function will read
     /// from the underlying [`tokio::io::AsyncReadExt`] until it does, an error occurs or no more data can be read (EOF).
     ///
-    /// If an EOF occurs and the specified amount could not be read, this function will return an [`ErrorKind::UnexpectedEof`].
-    ///
     /// This function does not consume data from the buffer, so subsequent calls to `peek` or `read` functions
     /// will still return the peeked data.
     ///
+    /// # Errors
+    ///
+    /// - If any error occurs while reading from the underlying [`tokio::io::AsyncReadExt`] it is returned
+    /// - If an EOF occurs and the specified amount could not be read, this function will return an [`ErrorKind::UnexpectedEof`].
+    ///
+    /// # Panics
+    ///
+    /// Will panic when attempting to read more bytes then `BUFFER_SIZE`
     pub async fn peek_exact(&mut self, amount: usize) -> Result<&[u8], MessageReadError> {
         self.fetch(amount, false).await
     }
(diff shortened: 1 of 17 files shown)


mavlink-core/src/bytes.rs
```diff
@@ -66,16 +78,25 @@ impl<'a> Bytes<'a> {
         val
     }
 
+    /// # Panics
+    ///
+    /// Will panic if not enough bytes remain for a `u16`
```
joaoantoniocardoso: [optional] Maybe specify x bytes here and elsewhere?

joaoantoniocardoso (approved):
Ty

mavlink-core/src/async_peek_reader.rs
```diff
@@ -84,10 +96,16 @@ impl<R: AsyncReadExt + Unpin, const BUFFER_SIZE: usize> AsyncPeekReader<R, BUFFE
...
     /// This function consumes the data from the buffer, unless an error occurs, in which case no data is consumed.
     ///
+    /// # Errors
+    ///
+    /// - If any error occurs while reading from the underlying [`tokio::io::AsyncReadExt`] it is returned
+    /// - If an EOF occurs before a byte could be read, this function will return an [`ErrorKind::UnexpectedEof`].
+    ///
+    /// # Panics
+    ///
+    /// Will always panic if `BUFFER_SIZE` is 0.
```
onur-ozkan: There is no such thing `BUFFER_SIZE` used in this function.

mavlink-core/src/async_peek_reader.rs
```diff
@@ -84,10 +96,16 @@ impl<R: AsyncReadExt + Unpin, const BUFFER_SIZE: usize> AsyncPeekReader<R, BUFFE
...
+    ///
+    /// - If any error occurs while reading from the underlying [`tokio::io::AsyncReadExt`] it is returned
+    /// - If an EOF occurs before a byte could be read, this function will return an [`ErrorKind::UnexpectedEof`].
+    ///
+    /// # Panics
+    ///
+    /// Will always panic if `BUFFER_SIZE` is 0.  
     pub async fn read_u8(&mut self) -> Result<u8, MessageReadError> {
         let buf = self.read_exact(1).await?;
         Ok(buf[0])
```
onur-ozkan, line 111: nit: This isn't infallible function, we don't have to risk the `buf[0]` panic here. We can use `get(0)` and return `MessageReadError` on `None` instead.

onur-ozkan:
Will you have any free space to finalize this PR? @pv42 

It would be quite cool to have this on the next release along with [Add wrapper for char arrays to access str conveniently #388](https://github.com/mavlink/rust-mavlink/pull/388).

mavlink-core/src/async_peek_reader.rs
pv42: `buf[0]` can not actually panic since the underling `fetch` returns either an `MessageReadError` or a slice of the specified len

onur-ozkan: How? `Ok(buf[0])` wouldn't be possible if `buf[0]` would return `MessageReadError`. I guess you mean `read_exact`?

pv42: Yeah `read_exact(1)` is just `fetch(1, true)`

onur-ozkan (approved):
Thanks!

PR #415, opened 2025-10-08
Improve string serde serialization/deserialization

patrickelectric:
This properly converts strings from inside or outside serde.


mavlink-core/src/utils.rs
```diff
@@ -66,3 +66,32 @@ mod tests {
...
+        D: Deserializer<'de>,
+    {
+        let s: String = Deserialize::deserialize(deserializer)?;
+        let mut buf = [0u8; N];
+        let bytes = s.as_bytes();
+        let len = bytes.len().min(N.saturating_sub(1));
+        buf[..len].copy_from_slice(&bytes[..len]);
+        buf[len] = 0;
+        Ok(buf)
+    }
```
pv42, line 96: This does not seem correct, at least it is not generally applicaple since some messages allow using none null terminated strings when the full buffer is filled with data.

patrickelectric: Yes, I just noticed it, I'm adding more tests to enforce the correct behavior.

Issue #401, opened 2025-09-19
Handling MAV_BOOL


azerupi:
Hi!

Upstream MAVLink has introduced the `MAV_BOOL` enum:

- https://github.com/mavlink/mavlink/pull/2242
- https://github.com/mavlink/mavlink/pull/2317

```xml
<enum name="MAV_BOOL" bitmask="true">
    <description>Enum used to indicate true or false (also: success or failure, enabled or disabled, active or inactive).</description>
    <entry value="0" name="MAV_BOOL_FALSE">
      <description>False.</description>
    </entry>
    <entry value="1" name="MAV_BOOL_TRUE">
      <description>True.</description>
    </entry>
</enum>
```

It is defined as a bit field and the generated Rust code is:
```rust
bitflags! { 
    # [cfg_attr (feature = "serde" , derive (Serialize , Deserialize))] 
    # [cfg_attr (feature = "arbitrary" , derive (Arbitrary))] 
    # [derive (Debug , Copy , Clone , PartialEq)] 
    # [doc = "Enum used to indicate true or false (also: success or failure, enabled or disabled, active or inactive)."] 
    pub struct MavBool : i8 { 
        # [doc = "False."] 
        const MAV_BOOL_FALSE = 0 ; 
        # [doc = "True."] 
        const MAV_BOOL_TRUE = 1 ; 
    } 
}
```

But then the enum is used in fields of different types:

- [In MAV_CMD params](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L1117)
- [As uint8_t field](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L6628)
- [As int8_t field](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L7051)

And the generated code fails to build because of operations between `i8` and `u8`. For example:

```
error[E0308]: mismatched types
     |
7686 |         __struct.capture_result = MavBool::from_bits(tmp & MavBool::all().bits()).ok_or(
     |                                                            ^^^^^^^^^^^^^^^^^^^^^ expected `i8`, found `u8`

error[E0277]: no implementation for `i8 & u8`
     |
7686 |         __struct.capture_result = MavBool::from_bits(tmp & MavBool::all().bits()).ok_or(
     |                                                          ^ no implementation for `i8 & u8`
     |
     = help: the trait `BitAnd<u8>` is not implemented for `i8`
[...]

PR #423, opened 2025-10-15
Support for MAV_BOOL

pv42:
Implementation of what I proposed here: https://github.com/mavlink/rust-mavlink/issues/401#issuecomment-3334395891
This does not actually bump the mavlink definitions so the MAV_BOOL is not actually added but rather support for it:
- add support for multiple representations for bitflags
- add special case `as_bool`

There were some other proposals in #401 so I am not sure if some other proposal is prefered/being worked on.

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -608,7 +608,7 @@ impl MavEnum {
             quote!()
         };
 
-        let special_impl = if self.name == "MavBool"
+        let mav_bool_impl = if self.name == "MavBool"
             && self
                 .entries
                 .iter()
@@ -662,7 +662,7 @@ impl MavEnum {
 
             impl #enum_name {
                 #const_default
-                #special_impl
+                #mav_bool_impl
             }
 
             impl Default for #enum_name {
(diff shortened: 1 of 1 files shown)


mavlink-bindgen/src/parser.rs
```diff
@@ -1226,6 +1254,26 @@ impl MavType {
...
     }
 
+    fn max_int_value(&self) -> u64 {
+        match self {
+            MavType::UInt8MavlinkVersion => u8::MAX as u64,
+            MavType::UInt8 => u8::MAX as u64,
+            MavType::UInt16 => u16::MAX as u64,
+            MavType::UInt32 => u32::MAX as u64,
+            MavType::UInt64 => u64::MAX,
+            MavType::Int8 => i8::MAX as u64,
```
onur-ozkan: `Int8`, `Char` and `CharArray` can be combined.

mavlink-bindgen/src/parser.rs
```diff
@@ -1226,6 +1254,26 @@ impl MavType {
...
+            MavType::Int16 => i16::MAX as u64,
+            MavType::Int32 => i32::MAX as u64,
+            MavType::Int64 => i64::MAX as u64,
+            MavType::Char => i8::MAX as u64,
+            // maximum precisly representable value minus 1 for float types
+            MavType::Float => 0xff_ffff,
+            MavType::Double => 0x1f_ffff_ffff_ffff,
```
onur-ozkan: Why not writing them as a human-readable way?

mavlink-bindgen/src/parser.rs
```diff
@@ -636,6 +662,7 @@ impl MavEnum {
 
             impl #enum_name {
                 #const_default
+                #special_impl
```
onur-ozkan: I think `special_impl` naming is not good here, it doesn't give the impression of `mav_bool`.

PR #426, opened 2025-10-21
Fix warnings in test and for certain dialects

pv42:


mavlink-bindgen/src/parser.rs
```diff
@@ -469,11 +465,19 @@ impl MavProfile {
...
+                    None
+                }
+            }
+        } else {
+            quote! {
+                fn target_system_id(&self) -> Option<u8> {
+                    match self {
+                        #(#arms)*
+                        _ => None,
+                    }
```
onur-ozkan: Can we ignore the warning with allowing it? The code became very ugly tbh.

mavlink-bindgen/src/parser.rs
```diff
@@ -491,11 +495,19 @@ impl MavProfile {
...
+                    None
+                }
+            }
+        } else {
+            quote! {
+                fn target_component_id(&self) -> Option<u8> {
+                    match self {
+                        #(#arms)*
+                        _ => None,
+                    }
```
onur-ozkan: Same here.

PR #428, opened 2025-10-27
refactor bytes module to be panic safe

onur-ozkan:
Makes bytes module to be panic safe, which is part of the internal implementation (not the public API).

### DIFF
--- a/mavlink-core/src/bytes.rs
+++ b/mavlink-core/src/bytes.rs
@@ -1,8 +1,27 @@
-use std::io;
-
 pub struct Bytes<'a> {
     data: &'a [u8],
     pos: usize,
+}
+
+/// Simple error types for the bytes module.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Error {
+    /// Attempted to to read more bytes than available.
+    NotEnoughBuffer { requested: usize, available: usize },
+}
+
+impl core::fmt::Display for Error {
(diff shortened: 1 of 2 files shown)


mavlink-core/src/bytes.rs
```diff
@@ -1,3 +1,5 @@
+use std::io;
```
joaoantoniocardoso: Hi, any use of `std` on the mavlink lib needs a special treatment to be compatible with embedded/no-std.

onur-ozkan: Yeah I realized that from the CI failure, will fix it shortly.

mavlink-core/src/bytes.rs
```diff
@@ -3,6 +3,27 @@ pub struct Bytes<'a> {
     pos: usize,
 }
 
+/// Simple error types for the bytes module.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Error {
+    /// Attempted to to read more bytes than available.
+    NotEnoughBuffer { requested: usize, available: usize },
```
joaoantoniocardoso, line 10: I wish other libraries did the same instead of returning UnexpectedEof

joaoantoniocardoso (approved):
Well done

mavlink-core/src/bytes.rs
onur-ozkan: Agreed. That would be a perfect follow-up work to this.

PR #430, opened 2025-11-01
remove duplicate bounds check in bytes

pv42:
With the newly introduced `NotEnoughBuffer` check_remaining is no longer required and the direct array index can be replaced with `.get(..).ok_or_else(|| ,,)`. This is a minor performance improvement since it eliminates the array bounds check that the `[index]` would perform. I measured about 0.5% better performance for parsing the messages from tests/log.tlog (without disk operations).

### DIFF
--- a/mavlink-core/src/bytes.rs
+++ b/mavlink-core/src/bytes.rs
@@ -10,6 +10,16 @@ pub enum Error {
     NotEnoughBuffer { requested: usize, available: usize },
 }
 
+impl Error {
+    #[inline]
+    fn not_enough_buffer(requested: usize, bytes: &Bytes) -> Self {
+        Self::NotEnoughBuffer {
+            requested,
+            available: bytes.remaining(),
+        }
+    }
+}
+
 impl core::fmt::Display for Error {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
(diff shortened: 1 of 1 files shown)


mavlink-core/src/bytes.rs
```diff
@@ -40,14 +40,10 @@ impl<'a> Bytes<'a> {
...
-            Ok(())
-        } else {
-            Err(Error::NotEnoughBuffer {
-                requested: count,
-                available: self.remaining(),
-            })
+    fn bounds_error(&self, requested: usize) -> Error {
+        Error::NotEnoughBuffer {
+            requested,
+            available: self.remaining(),
```
onur-ozkan: You can implement `fn not_enough_buffer` under `impl Error` directly.

pv42: Could you maybe clarify ? I can't move `fn bounds_error` itself since it has to call `self.remaining()` and creating a function just to wrap `Error::NotEnoughBuffer { requested, available}` seems like unnecesarry bloat since it would do nothing extra and only be used in this one place.

onur-ozkan: You can pass the remaining value the same way you pass the requested value. The bloating is already happening with the current approach and it's not just bloat but also an uncommon way to return errors. If you want to avoid bloat, you should return the error directly instead of having to call a function.

pv42: I moved `bounds_error` into `Error::not_enough_buffer`

Issue #393, opened 2025-09-09
[Bug] Fields without description are not generated


pv42:
Was adding check to enforce that the message field count is 1<=count<=64 when I noticed that this message seems to be generated incorrectly:
`cubepilot.xml`: `CUBEPILOT_RAW_RC_DATA` 

XML:
```xml
<message id="50001" name="CUBEPILOT_RAW_RC">
    <description>Raw RC Data</description>
    <field type="uint8_t[32]" name="rc_raw"/>
</message>
```
generated message in [doc](https://docs.rs/mavlink/0.15.0/mavlink/cubepilot/struct.CUBEPILOT_RAW_RC_DATA.html)
Source:
```rs
// #[...]
pub struct CUBEPILOT_RAW_RC_DATA {}
```

Expected generation is something like
```rs
pub struct CUBEPILOT_RAW_RC_DATA {
    pub rc_raw: [u8; 32],
}
```
Bug exists at least since 11.2 (earler version do not have this message), but 10.x is likely also affected.

While most message have a description this is not mandatory [doc](https://mavlink.io/en/guide/define_xml_element.html#creating-a-message).

pv42:
Only other occurrence in official dialect definitions seems to be `HERELINK_TELEM`.

PR #437, opened 2025-11-28
Documentation for MAV_CMD variants

pv42:
MAV_CMD should get better support that the current method of manually having to fill the paramters fields. This adds the XML parsing part and used it for documentation but can later be used for codegen too. 
Also simplify the XML parsing by treating self closing tags as an start and end tag, the self closing tag handling had just copied code anyways. This should also avoid some bugs like #393.

Example of new doc: 
grafik

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -2442,4 +2442,123 @@ mod tests {
         // Should panic due to no fields
         msg.validate_field_count();
     }
+
+    #[test]
+    fn test_fmt_mav_param_values() {
+        let enum_param = MavParam {
+            enum_used: Some("ENUM_NAME".to_string()),
+            ..Default::default()
+        };
+        assert_eq!(enum_param.format_valid_values(), "[`ENUM_NAME`]");
+
+        let reserved_param = MavParam {
+            reserved: true,
+            default: Some(f32::NAN),
+            ..Default::default()
+        };
+        assert_eq!(reserved_param.format_valid_values(), "Reserved (use NaN)");
+
+        let unrestricted_param = MavParam::default();
+        assert_eq!(unrestricted_param.format_valid_values(), "");
+
+        let int_param = MavParam {
+            increment: Some(1.0),
+            ..Default::default()
+        };
+        assert_eq!(int_param.format_valid_values(), "Multiples of 1");
+
+        let pos_param = MavParam {
+            min_value: Some(0.0),
+            ..Default::default()
+        };
+        assert_eq!(pos_param.format_valid_values(), "&ge; 0");
+
+        let max_param = MavParam {
+            max_value: Some(5.5),
+            ..Default::default()
+        };
+        assert_eq!(max_param.format_valid_values(), "&le; 5.5");
+
+        let pos_int_param = MavParam {
+            min_value: Some(0.0),
+            increment: Some(1.0),
+            ..Default::default()
+        };
+        assert_eq!(pos_int_param.format_valid_values(), "0, 1, ..");
+
+        let max_inc_param = MavParam {
+            increment: Some(1.0),
+            max_value: Some(360.0),
+            ..Default::default()
+        };
+        assert_eq!(max_inc_param.format_valid_values(), ".., 359, 360");
+
+        let range_param = MavParam {
+            min_value: Some(0.0),
+            max_value: Some(10.0),
+            ..Default::default()
+        };
+        assert_eq!(range_param.format_valid_values(), "0 .. 10");
+
+        let int_range_param = MavParam {
+            min_value: Some(0.0),
+            max_value: Some(10.0),
+            increment: Some(1.0),
+            ..Default::default()
+        };
+        assert_eq!(int_range_param.format_valid_values(), "0, 1, .. , 10");
+
+        let close_inc_range_param = MavParam {
+            min_value: Some(-2.0),
+            max_value: Some(2.0),
+            increment: Some(2.0),
+            ..Default::default()
+        };
+        assert_eq!(close_inc_range_param.format_valid_values(), "-2, 0, 2");
+
+        let bin_range_param = MavParam {
+            min_value: Some(0.0),
+            max_value: Some(1.0),
+            increment: Some(1.0),
+            ..Default::default()
+        };
+        assert_eq!(bin_range_param.format_valid_values(), "0, 1");
+    }
+
+    #[test]
+    fn test_emit_doc_row() {
+        let param = MavParam {
+            index: 3,
+            label: Some("test param".to_string()),
+            min_value: Some(0.0),
+            units: Some("m/s".to_string()),
+            ..Default::default()
+        };
+        // test with all variations of columns
+        assert_eq!(
+            param.emit_doc_row(false, false).to_string(),
+            quote! {#[doc = "| 3 (test param)|             |"]}.to_string()
+        );
+        assert_eq!(
+            param.emit_doc_row(false, true).to_string(),
+            quote! {#[doc = "| 3 (test param)|             | m/s |"]}.to_string()
+        );
+        assert_eq!(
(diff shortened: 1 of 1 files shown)


mavlink-bindgen/src/parser.rs
```diff
@@ -713,6 +717,108 @@ impl MavEnumEntry {
...
+                }
+            }
+            (Some(min), Some(max), None) => format!("{min} .. {max}"),
+            (Some(min), None, Some(inc)) => format!("{}, {}, .. ", min, min + inc),
+            (Some(min), None, None) => format!("&ge; {min}"),
+            (None, Some(max), Some(inc)) => format!(".. , {}, {}", max - inc, max),
+            (None, Some(max), None) => format!("&le; {max}"),
+            (None, None, Some(inc)) => format!("Multiples of {inc}"),
+            (None, None, _) => String::new(),
+        }
```
onur-ozkan: This is very hard to follow and debug. Also, moving lines would change the behavior significantly (which is dangerous). Can we somehow re-write this in a better and safer way?

mavlink-bindgen/src/parser.rs
```diff
@@ -1886,6 +1961,44 @@ pub fn parse_profile(
...
+                                if let Some(max) = param_max_value {
+                                    assert!(
+                                        min <= max,
+                                        "param minValue must not be greater then maxValue"
+                                    );
+                                }
+                            }
+                            params[param_index - 1] = MavParam {
+                                index: param_index,
+                                description: param_description.unwrap_or_default(),
```
onur-ozkan: Why not making `description` `Option<String` as well?

pv42: Yes

pv42:
I have added a snapshot test but it currently still suffers from #439.

mavlink-bindgen/src/parser.rs
pv42: I have improved this, the `(None, None, _) =>` is gone and the whole function does now don't have early returns anymore, but since our MSRV does not have if let chains yet it is still not ideal. Since there are 10 different cases i think a big match is still better then a giant if tree, but I reordered them a bit so i hopefully becomes more clear.

mavlink-bindgen/src/parser.rs
```diff
@@ -1907,6 +1979,43 @@ pub fn parse_profile(
                             deprecated.as_mut().unwrap().note = Some(t);
                         }
                     }
+                    Some(&MavXmlElement::Param) => {
+                        if let Some(params) = entry.params.as_mut() {
+                            // Some messages can jump between values, like:
+                            // 1, 2, 7
```
onur-ozkan: No need to wrap the comment here (unless rustfmt is being angry about it)

```suggestion
                            // Some messages can jump between values, like: 1, 2, 7
```

mavlink-bindgen/src/parser.rs
```diff
@@ -710,6 +714,108 @@ impl MavEnumEntry {
...
+                #[doc = ""]
+                #[doc = "# Parameters"]
+                #[doc = ""]
+                #[doc = #table_header]
+                #[doc = #table_hl]
+                #(#lines)*
+            }
+        } else {
+            quote!()
+        }
```
onur-ozkan, line 751: You could do something like:

```
let Some(params) = &self.params else {
    return quote!();
}
```

and write the rest of the code with better readibility.

mavlink-bindgen/src/parser.rs
```diff
@@ -710,6 +714,108 @@ impl MavEnumEntry {
...
+    pub index: usize,
+    pub description: Option<String>,
+    pub label: Option<String>,
+    pub units: Option<String>,
+    pub enum_used: Option<String>,
+    pub increment: Option<f32>,
+    pub min_value: Option<f32>,
+    pub max_value: Option<f32>,
+    pub reserved: bool,
+    pub default: Option<f32>,
```
onur-ozkan, line 767: Better ordering:

```suggestion
    pub index: usize,
    pub reserved: bool,
    pub description: Option<String>,
    pub label: Option<String>,
    pub units: Option<String>,
    pub enum_used: Option<String>,
    pub increment: Option<f32>,
    pub min_value: Option<f32>,
    pub max_value: Option<f32>,
    pub default: Option<f32>,
```

mavlink-bindgen/src/parser.rs
```diff
@@ -710,6 +714,108 @@ impl MavEnumEntry {
...
+}
+
+impl MavParam {
+    fn format_valid_values(&self) -> String {
+        if self.reserved && self.default.is_some() {
+            format!("Reserved (use {})", self.default.unwrap())
```
onur-ozkan: Clippy is not happy here (I also think it looks some annoying).

Can you try this?

```rs
if let (true, Some(default)) = (self.reserved, self.default)
```

mavlink-bindgen/src/parser.rs
```diff
@@ -710,6 +714,108 @@ impl MavEnumEntry {
...
+                    }
+                }
+                (Some(min), Some(max), None) => format!("{min} .. {max}"),
+                (Some(min), None, Some(inc)) => format!("{}, {}, .. ", min, min + inc),
+                (None, Some(max), Some(inc)) => format!(".. , {}, {}", max - inc, max),
+                (Some(min), None, None) => format!("&ge; {min}"),
+                (None, Some(max), None) => format!("&le; {max}"),
+                (None, None, Some(inc)) => format!("Multiples of {inc}"),
+                (None, None, None) => String::new(),
+            }
```
onur-ozkan: If we could do this in a separate function, we would have a complete test coverage on it. Without complete coverage, any change on this block of code will take very long time to review or risk some bugs on merge.

mavlink-bindgen/src/parser.rs
```diff
@@ -710,6 +714,108 @@ impl MavEnumEntry {
...
+        if value_range_col {
+            let range = self.format_valid_values();
+            line += &format!(" {range} |");
+        }
+        if units_row {
+            let units = self.units.clone().unwrap_or_default();
+            line += &format!(" {units} |");
+        }
+        quote! {#[doc = #line]}
+    }
```
onur-ozkan: A strong coverage would be very good for the reasons I explained above.

mavlink-bindgen/src/parser.rs
```diff
@@ -1907,6 +1979,43 @@ pub fn parse_profile(
...
+                                });
+                            }
+                            if let Some(min) = param_min_value {
+                                if let Some(max) = param_max_value {
+                                    assert!(
+                                        min <= max,
+                                        "param minValue must not be greater then maxValue"
+                                    );
+                                }
+                            }
```
onur-ozkan: Cleaner way:

```rs
param_min_value.zip(param_max_value).map(|(min, max)| {
    assert!(min <= max, "param minValue must not be greater than maxValue");
});
```

mavlink-bindgen/src/parser.rs
pv42: Ordering is based on [the XML doc](https://mavlink.io/en/guide/xml_schema.html#param). Since `reserved` and `default` are related they should be next to each other. `reserved` is theoretuically `Option<bool>` but has a distinct default of false hence I used `bool` instead and initialize with the default value.

mavlink-bindgen/src/parser.rs
pv42: Clippy didn't quite like the suggestion hence I modified it slightly to
```rs
if let Some((min, max)) = param_min_value.zip(param_max_value) { 
  assert!(min <= max, "param minValue must not be greater than maxValue")
}
```

mavlink-bindgen/src/parser.rs
onur-ozkan: That's fine. I was suggesting a type-based ordering but it's nothing serious.

PR #438, opened 2025-11-28
New method created to expose socket local_addr

joaofrf:
Implemented a method to get the local_addr of udp and tcp sockets

This is particularly helpful when doing something like this:

`
let connection = mavlink_core::connect_async::<MavMessage>("udpin:127.0.0.1:0")
`

where it will generate the socket with an unknown port.
Using the new method we can retrieve what port was used and send it to who ever needs to send information to that port.

### DIFF
--- a/mavlink-core/src/async_connection/direct_serial.rs
+++ b/mavlink-core/src/async_connection/direct_serial.rs
@@ -124,13 +124,6 @@ impl<M: Message + Sync + Send> AsyncMavConnection<M> for AsyncSerialConnection {
         self.recv_any_version
     }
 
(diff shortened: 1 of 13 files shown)


onur-ozkan (changes requested):
It's not good to implement socket related function to every type of mavlink connection.

Instead of that, we can replace `address: String` with `address: Socket` in `UdpConfig` and `TcpConfig`. That way the following should be possible:

```rs
let socket = UdpSocket::bind("0.0.0.0:0")?;
let _port = socket.local_addr()?;
let connection = UdpConfig::new(socket, ...).connect()?; 
```

PR #441, opened 2025-12-07
fix(427): arbitrary handles utf-8 bytes with truncation

### LINKED ISSUE #427: `arbitrary` generated `CharArray`s are incompateble with `serde`
`arbitrary` generates randomly filled u8 buffer but when serializing `nulstr::serialize` expects a valid utf8 string causing a error when attempting to serialize.
e.g. invalid utf-8 sequence of 1 bytes from index 1 on 
```
CHANGE_OPERATOR_CONTROL(CHANGE_OPERATOR_CONTROL_DATA {
 target_system: 94,
 control_request: 79,
 version: 180,
 passkey: CharArray {
  data: [35, 139, 164, 41, 176, 162, 155, 2, 145, 166, 186, 11, 14, 153, 127, 133, 19, 23, 30, 218, 106, 221, 84, 77, 249],
  str_len: 25
 }
})
```

bengtlofgren:
# Fix #427: Arbitrary handles UTF-8 bytes with truncation

## Summary

Fixes the `Arbitrary` implementation for `CharArray<N>` to properly handle UTF-8 encoded strings instead of filling with random bytes. This was inspired from the disccusion [here](https://github.com/mavlink/rust-mavlink/issues/427#issuecomment-3498246724), attempting to solve #427 .

## Problem

The previous implementation used `u.fill_buffer(&mut data)` which filled the array with arbitrary bytes. This could produce invalid UTF-8 sequences, causing issues when the `CharArray` was expected to contain valid string data.

## Solution

- Generate only u8's from the readable ascii characters

This ensures that fuzz testing with the `arbitrary` feature produces valid UTF-8 content that won't cause unexpected behavior.

## Changes

- `mavlink-core/src/types.rs`: Updated `Arbitrary` impl for `CharArray<N>` with UTF-8 aware truncation

### DIFF
--- a/mavlink-core/src/types.rs
+++ b/mavlink-core/src/types.rs
@@ -137,7 +137,12 @@ impl<'de, const N: usize> Deserialize<'de> for CharArray<N> {
 impl<'a, const N: usize> Arbitrary<'a> for CharArray<N> {
     fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
         let mut data = [0u8; N];
-        u.fill_buffer(&mut data)?;
+
+        for b in &mut data {
+            // Take a char from the printable ASCII range.
+            *b = u.int_in_range(32..=126)?;
+        }
+
         Ok(CharArray::new(data))
     }
 }


mavlink-core/src/types.rs
```diff
@@ -136,8 +136,22 @@ impl<'de, const N: usize> Deserialize<'de> for CharArray<N> {
...
+            // Find the last char boundary that fits within N bytes
+            let mut end = N;
+            while end > 0 && !s.is_char_boundary(end) {
+                end -= 1;
+            }
+            &s[..end]
+        };
+
+        data[..truncated.len()].copy_from_slice(truncated.as_bytes());
         Ok(CharArray::new(data))
```
onur-ozkan, line 146: Thank you for your PR.

This is too much and heavy (e..g, the heap allocation) for a simple task like this.

How about something like this:

```rs
let mut data = [0u8; N];

for b in &mut data {
    // Take a char from the printable ASCII range.
    *b = u.int_in_range(32..=126)?; 
}

Ok(CharArray::new(data))
```

bengtlofgren: You're right, I didn't consider heap was important for fuzzing, but this is so much cleaner

onur-ozkan:
Thanks! Btw, the PR description is outdated. It documents the first attempt.

PR #447, opened 2025-12-20
remove RustDefault

danieleades:


danieleades:
> If I understood it correctly, the proposal is removing a feature to make the code compatible with an outdated building tool. Questions: (1) could this feature be useful for anyone else? (2) Why not disable the feature?
> 
> Another option is to custom-make an image compatible with cross like [these](https://hub.docker.com/r/joaoantoniocardoso/cross-rs/tags) bullseye with gstreamer images that we use to cross-build Mavlink Camera Manager.

not exactly. It's removing a workaround which was needed for older versions of rust which is no longer needed. Except it *is* needed in `cross` environments because the images are so old.

There is no impact to users of the library from removing this workaround, but it simplifies the codebase and improves maintainability.

increasing the complexity of the CI in order to decrease the simplicity of the code doesn't seem particularly useful. I recommend waiting until `cross` makes their next release before merging this.

PR #454, opened 2026-01-16
Bump MAVLink version

tbolin:
Bump the MAVLink version to the latest as of 2026-01-16.

Also:
* Add support for the `superseded` tag , which is required for the new message definitions.
* Add feature gates and documentation for *marsh* and *stemstudios* dialects.

### DIFF
--- a/mavlink-bindgen/tests/snapshots/e2e_snapshots__mav_bool.xml@mav_bool.rs.snap
+++ b/mavlink-bindgen/tests/snapshots/e2e_snapshots__mav_bool.xml@mav_bool.rs.snap
@@ -28,7 +28,7 @@ bitflags! { # [cfg_attr (feature = "ts" , derive (TS))] # [cfg_attr (feature = "
 impl MavBool {
     pub const DEFAULT: Self = Self::MAV_BOOL_FALSE;
     pub fn as_bool(&self) -> bool {
-        *self == Self::MAV_BOOL_TRUE
+        self.contains(Self::MAV_BOOL_TRUE)
     }
 }
 impl Default for MavBool {


tbolin:
I have added feature gates and documentation for the new dialects. However, `cargo test --features arbitrary,standard` fails.
I think the problem is that `common` defines but does not actually use the `MavBool` enum, so the parser defines it as a normal enum and not with bitflags, and then 
```
    pub fn as_bool(&self) -> bool {
        self.contains(Self::MAV_BOOL_TRUE)
    }
```
is invalid, since `self.contains` doesn't exist.

mavlink/Cargo.toml
```diff
@@ -80,7 +82,9 @@ all-dialects = [
     "cubepilot",
     "storm32",
     "csairlink",
-    "loweheiser"
+    "loweheiser",
+    "marsh",
+    "stemstudios"
```
onur-ozkan: Please add "," to the end so we don't cause diff to this line when adding new dialect.

```suggestion
    "stemstudios",
```

mavlink/src/lib.rs
```diff
@@ -35,12 +35,14 @@
 //! - `development`
 //! - `icarous`
 //! - `loweheiser`
+//! - `mars`
```
onur-ozkan: Typo

```suggestion
//! - `marsh`
```

mavlink-bindgen/src/parser.rs
```diff
@@ -638,7 +639,7 @@ impl MavEnum {
         {
             quote!(
                 pub fn as_bool(&self) -> bool {
-                    self.contains(Self::MAV_BOOL_TRUE)
+                    *self == Self::MAV_BOOL_TRUE
```
onur-ozkan: This is not right as `self` implements bitflags.

tbolin: I need some input on how to solve this.
The problem is that `self` does not implement bitflags if the dialect defines MAV_BOOL but does not use it in any messages (e.g. the standard dialect).

I think the best solution would be to default to either u8 or i8 if the type of MAV_BOOL can't be deduced from how it is used in the dialect. Then MAV_BOOL would always implement bitflags and `self.contains(Self::MAV_BOOL_TRUE)` would always be valid.

The only way I could see that this could cause problems is if a dialect that currently defines but not use MAV_BOOL starts using MAV_BOOL in a way that does not agree with the default. 
However, I'm not sure if the current solution (not implementing bitflags at all) would be any better in that case.

I'm also open to other suggestions.

mavlink/Cargo.toml
tbolin: fixed

mavlink-bindgen/src/parser.rs
tbolin: I no one has any comments, I'm going to start implementing this solution.

onur-ozkan: Sorry for the delay. I have had a lot going on these days. I will look into it over the weekend -most likely by the end of it- and propose a solution (if there is any better way).

onur-ozkan: I just created https://github.com/mavlink/rust-mavlink/pull/459 which handles this problem in a consistent and stable way. You can rebase this PR once it's merged (probably within a couple of days).

tbolin: Rebased on master with https://github.com/mavlink/rust-mavlink/pull/459

PR #480, opened 2026-03-25
Doc formating

pv42:
Some doc improvements mostly based on the 2 lints bellow. Additionally the `parser::generate()` function doc had references to things that do not apply anymore (message conversion between dialects, protobuf (?)). There are no functional changes.
### clippy::doc_link_code
- merge code fragments that were next to each other using `<code>` instead of multiple ``
- enable lint
### clippy::doc_paragraphs_missing_punctuation
- apply lint for all functions, but skip over fields, structs, etc. unless the doc was already a sentence
- adjusted comment for function to build proper sentences (when required)
- don't enble the lint since it triggers in cases where punctuation would not make sence e.g. struct fields like `system_id`


mavlink-core/src/lib.rs
```diff
@@ -558,12 +558,12 @@ pub fn read_versioned_msg_signed<M: Message, R: Read>(
...
 /// Asynchronously read and parse a MAVLinkMessageRaw of the specified version from a [`AsyncPeekReader`] with signing support.
 ///
-/// When using [`ReadVersion::Single`]`(`[`MavlinkVersion::V1`]`)` signing is ignored.
+/// When using <code>[ReadVersion::Single]([MavlinkVersion::V1])</code> signing is ignored.
 /// When using [`ReadVersion::Any`] MAVlink 1 messages are treated as unsigned.
 ///
 /// # Errors
 ///
-/// See [`read_` function error documentation](crate#read-errors)
+/// See [`read_` function error documentation](crate#read-errors).
```
onur-ozkan, line 566: I don't think this change is necessary. It doesn't add much value and it overlaps with the existing history, which may make it harder for someone to find the commit where the change was originally introduced.

Same applies to majority of this PR. I only see couple of lines where the change actually makes sense.

Cargo.toml
```diff
@@ -22,3 +22,4 @@ semicolon_if_nothing_returned = "warn"
 unnecessary_semicolon = "warn"
 uninlined_format_args = "warn"
 use_self = "warn"
+doc_link_code = "warn"
```
onur-ozkan, line 25: I am not sure this is a good lint. :/ I am personally more happy without the html tags being added to the repository.

I will leave this to other maintainters.

pv42: Also not a fan of the `<code>` but it's impossible to achieve the same with ``. The previous creates multiple unconnected inline code blocks.

PR #481, opened 2026-03-25
Resolve some todos in code

pv42:
There are some `//TODO`s in the code that can be removed 

`//TODO verify this is no longer necessary since we're supporting both mavlink1 and mavlink2`; `//TODO verify that id_width of u8 is OK even in mavlink v1`:

- We have an explicit error `MessageWriteError::MAVLink2Only` that check for such cases 
- There is no functional way to generate only MAVLink1 messages right now

`// TODO: these lints are produced on a couple of cubepilot messages (..)`

- the cubepilot dialect is fixed
- unclosed tags should be caught by quickxml, unless selfclosing tags are meant, with are now supported
- added an assert that fails when a message would have a 0 bytes payload (this is already forbidden since messages must have at least one field)

`//TODO check` in MavFrame::ser()

- use a checked int conversion instead, since the function can already panic on incorrect usage that should be fine
- using the uncheck cast would result in incorect bytes streams when it overflows

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -1005,10 +1005,9 @@ impl MavMessage {
         let extra_crc = extra_crc(self);
         let (name_types, payload_encoded_len) = self.emit_name_types();
         assert!(
-            payload_encoded_len <= 255,
-            "maximum payload length is 255 bytes"
+            (1..=255).contains(&payload_encoded_len),
+            "payload length must be between 1 and 255 bytes"
         );
-        assert!(payload_encoded_len > 0, "payload may not be empty");
 
         let deser_vars = self.emit_deserialize_vars();
         let serialize_vars = self.emit_serialize_vars();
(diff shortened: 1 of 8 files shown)


mavlink-bindgen/src/parser.rs
```diff
@@ -1028,6 +1008,7 @@ impl MavMessage {
             payload_encoded_len <= 255,
             "maximum payload length is 255 bytes"
         );
+        assert!(payload_encoded_len > 0, "payload may not be empty");
```
onur-ozkan: These 2 can be combined.

mavlink-core/src/lib.rs
```diff
@@ -301,7 +301,8 @@ impl<M: Message> MavFrame<M> {
     ///
     /// # Panics
     ///
-    /// Will panic if frame does not fit in the provided buffer.
+    /// - if the frame does not fit in the provided buffer
+    /// - when attempting to serialize a message with an id greater then 255 with MAVLink 1
```
onur-ozkan:
```suggestion
    /// - If the frame does not fit in the provided buffer
    /// - When attempting to serialize a message with an id greater then 255 with MAVLink 1
```

PR #488, opened 2026-04-16
Avoid oversized payload parsing buffer allocation when not required

pv42:
performance difference is ~5%
tested on arbitrary messages from with ids 0-255 (if they exist)
```
before:
mavlink v1:       1,983.18 ns/iter (+/- 31.22)
mavlink v2:       1,994.77 ns/iter (+/- 83.17)
after:
mavlink v1:       1,893.43 ns/iter (+/- 186.79)
mavlink v2:       1,903.78 ns/iter (+/- 89.57)
```
test code:
```rs
bencher.iter(|| {
    for (id, buf) in &msg_bufs {
        std::hint::black_box(MavMessage::parse(version, *id, buf)).unwrap();
    }
})
```

### DIFF
--- a/mavlink-bindgen/src/parser.rs
+++ b/mavlink-bindgen/src/parser.rs
@@ -973,9 +973,10 @@ impl MavMessage {
             quote! {
                 let avail_len = __input.len();
 
-                let mut payload_buf  = [0; Self::ENCODED_LEN];
+                let mut payload_buf;
                 let mut buf = if avail_len < Self::ENCODED_LEN {
                     //copy available bytes into an oversized buffer filled with zeros
+                    payload_buf = [0; Self::ENCODED_LEN];
                     payload_buf[0..avail_len].copy_from_slice(__input);
                     Bytes::new(&payload_buf)
                 } else {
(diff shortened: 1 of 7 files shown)


mavlink-bindgen/src/parser.rs
```diff
@@ -953,9 +973,10 @@ impl MavMessage {
             quote! {
                 let avail_len = __input.len();
 
-                let mut payload_buf  = [0; Self::ENCODED_LEN];
+                let mut payload_buf;
                 let mut buf = if avail_len < Self::ENCODED_LEN {
                     //copy available bytes into an oversized buffer filled with zeros
+                    payload_buf = [0; Self::ENCODED_LEN];
```
onur-ozkan, line 979: This can simply be:

```suggestion
                let mut buf = if avail_len < Self::ENCODED_LEN {
                    //copy available bytes into an oversized buffer filled with zeros
                    let payload_buf = [0; Self::ENCODED_LEN];
```

I also wonder why do we need to allocate more than we need. The code does:

```
payload_buf[0..avail_len].copy_from_slice(__input);
```

which makes `[avail_len..Self::ENCODED_LEN]` wasted. The else block seems much more straightforward. I think it's worth looking into whether we can do that unconditionally.

onur-ozkan (changes requested):

mavlink-bindgen/src/parser.rs
pv42: The suggestion does not work since `payload_buf` needs to stay in scope for the actual parsing.

The whole buffer part is done since Mavlink 2 messages remove all zero bytes at the end of the payload.

I briefly tried modifying `Bytes` to allow returning more bytes regardless of its underlying buffer by generating the zero bytes itself when required but it seemed a lot slower.

onur-ozkan: Ah yes, the lifetime problem. Thanks.

onur-ozkan (approved):

PR #491, opened 2026-05-05
remove deprecated read_raw functions

pv42:
Since there is a breaking release upcoming these can be removed now.


onur-ozkan (approved):
Makes sense, thanks for tracking these!

PR #493, opened 2026-05-25
refactor(bytes_mut): replace unsafe copy with safe copy_from_slice

onur-ozkan:
Use `copy_from_slice` in `BytesMut::put_slice` instead of manually calling `copy_nonoverlapping`. The existing bounds check already ensures the destination range is valid, so the safe slice API is equivalent and removes unnecessary unsafe code.

### DIFF
--- a/mavlink-core/src/bytes_mut.rs
+++ b/mavlink-core/src/bytes_mut.rs
@@ -39,7 +39,7 @@ impl<'a> BytesMut<'a> {
     pub fn put_slice(&mut self, src: &[u8]) {
         self.check_remaining(src.len());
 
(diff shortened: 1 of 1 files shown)


joaoantoniocardoso (approved):
Thanks, this is indeed better.

optionally, the code could be reduced if all the put_x_le functions use put_slice: `self.put_slice(&val.to_le_bytes())`. Not sure it's good for performance, though.

onur-ozkan:
> Thanks, this is indeed better.
> 
> optionally, the code could be reduced if all the put_x_le functions use put_slice: `self.put_slice(&val.to_le_bytes())`. Not sure it's good for performance, though.

They have different size handling logic.

PR #496, opened 2026-05-29
adjust bitflag size check, add SerializationError

### LINKED ISSUE #495: bindgen fails on GIMBAL_DEVICE_CAP_FLAGS on latest mavlink definitions
After updating the mavlink submodule, `cargo build` fails with:

```sh
error: failed to run custom build command for `mavlink v0.18.0 (rust-mavlink/mavlink)`

Caused by:
  process didn't exit successfully: `rust-mavlink/target/debug/build/mavlink-3ef70f176d0bae59/build-script-main` (exit status: 101)
  --- stderr

  thread 'main' (34784) panicked at mavlink-bindgen/src/parser.rs:90:33:
  bitflag enum field cap_flags of GIMBAL_DEVICE_INFORMATION must be able to fit all possible values for GimbalDeviceCapFlags
```

This is because bindgen currently has this invariant:

> every bitmask enum value must fit into every field that references that enum

which is wrong for enums shared by legacy fields and wider extension fields.

e.g. `GIMBAL_DEVICE_INFORMATION` has:

```xml
<field type="uint16_t" name="cap_flags" enum="GIMBAL_DEVICE_CAP_FLAGS">
<field type="uint32_t" name="cap_flags2" enum="GIMBAL_DEVICE_CAP_FLAGS" invalid="0">
```

Because this contains values above `u16`, parser panics on the legacy `cap_flags` field.

Blocker for https://github.com/mavlink/rust-mavlink/issues/490

pv42:
Fixes #495

Since all `ser` and related function return a result now this is very much breaking. 
Also turned the panic on too small buffer size into an error, same should eventually also be done for the case of message ids over 255 if tried to use them with Mavlink1.

I would also like to add a test that checks that the error is actually generated, but that needs the updated mavlink definitions.

### DIFF
--- a/mavlink-bindgen/tests/definitions/oversized_enum.xml
+++ b/mavlink-bindgen/tests/definitions/oversized_enum.xml
@@ -1,0 +1,19 @@
+<mavlink>
+  <enums>
+    <enum name="GIMBAL_DEVICE_CAP_FLAGS" bitmask="true">
+      <entry value="1" name="GIMBAL_DEVICE_CAP_FLAGS_HAS_RETRACT">
+        <description>Gimbal device supports a retracted position.</description>
+      </entry>
+      <entry value="131072" name="GIMBAL_DEVICE_CAP_FLAGS_CAN_POINT_LOCATION_GLOBAL">
+        <description>Gimbal device supports to point to a global latitude, longitude, altitude position.</description>
+      </entry>
+    </enum>
+  </enums>
+  <messages>
+    <message id="283" name="GIMBAL_DEVICE_INFORMATION">
+      <description>Information about a low level gimbal. This message should be requested by the gimbal manager or a ground station using MAV_CMD_REQUEST_MESSAGE. The maximum angles and rates are the limits by hardware. However, the limits by software used are likely different/smaller and dependent on mode/settings/etc..</description>
+      <field type="uint16_t" name="cap_flags" enum="GIMBAL_DEVICE_CAP_FLAGS">Bitmap of gimbal capability flags.</field>
+      <field type="uint32_t" name="cap_flags2" enum="GIMBAL_DEVICE_CAP_FLAGS" invalid="0">Extended bitmap of gimbal capability flags (32 bit). For backwards compatibility, the lower 16 bits should also be set in cap_flags. Ground stations should prefer this field if non-zero.</field>
+    </message>
+  </messages>
+</mavlink>   
--- a/mavlink-bindgen/tests/definitions/oversized_enum.xml
+++ b/mavlink-bindgen/tests/definitions/oversized_enum.xml
@@ -16,4 +16,5 @@
       <field type="uint32_t" name="cap_flags2" enum="GIMBAL_DEVICE_CAP_FLAGS" invalid="0">Extended bitmap of gimbal capability flags (32 bit). For backwards compatibility, the lower 16 bits should also be set in cap_flags. Ground stations should prefer this field if non-zero.</field>
     </message>
   </messages>
-</mavlink>   
+</mavlink>
+ 
--- a/mavlink-core/src/lib.rs
+++ b/mavlink-core/src/lib.rs
@@ -101,7 +101,7 @@ use peek_reader::PeekReader;
 
 use crate::{
     bytes::Bytes,
-    error::{MessageReadError, MessageWriteError, ParserError},
+    error::{MessageReadError, MessageWriteError, ParserError, SerializationError},
 };
 
 use crc_any::CRCu16;
@@ -192,10 +192,10 @@ where
 
     /// Serialize **Message** into byte slice and return count of bytes written
     ///
-    /// # Panics
+    /// # Errors
     ///
-    /// Will panic if the buffer provided is to small to store this message
-    fn ser(&self, version: MavlinkVersion, bytes: &mut [u8]) -> usize;
+    /// Will return an error if the buffer provided is to small to store this message or when a bitflag value has an invalid value for a field
+    fn ser(&self, version: MavlinkVersion, bytes: &mut [u8]) -> Result<usize, SerializationError>;
 
     /// Parse a Message from its message id and payload bytes
     ///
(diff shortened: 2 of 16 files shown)


mavlink-core/src/lib.rs
```diff
@@ -226,10 +226,11 @@ pub trait MessageData: Sized {
...
     const ENCODED_LEN: usize;
 
-    /// # Panics
+    /// # Errors
     ///
-    /// Will panic if the buffer provided is to small to hold the full message payload of the implementing message type
-    fn ser(&self, version: MavlinkVersion, payload: &mut [u8]) -> usize;
+    /// Will return an error if the buffer provided is to small to hold the full message payload of the implementing message type or when a bitflag value has an invalid value for a field
+    fn ser(&self, version: MavlinkVersion, payload: &mut [u8])
+    -> Result<usize, SerializationError>;
```
onur-ozkan, line 233: Hmm. I didn't know that we were going to change core runtime interface from infallible to fallible.

My proposal is just fail bindgen on undersized types and leave the runtime part infallible.

Also, I think `BytesMut` would already handle some of the buffer sizing problems, am I wrong? If it's not doing that already then we need to improve `BytesMut` to do so.

mavlink-core/src/lib.rs
```diff
@@ -301,14 +302,19 @@ impl<M: Message> MavFrame<M> {
     ///
     /// # Panics
     ///
-    /// - If the frame does not fit in the provided buffer
```
onur-ozkan, line 304: `BytesMut` covers such cases and panics.

mavlink-core/src/lib.rs
pv42: > My proposal is just fail bindgen on undersized types and leave the runtime part infallible.

How would you propose handle someone trying to serialize an invalid bitflag value then ? We could silently slice it of and just mention in the doc that which fields are potentially do that.

> Also, I think BytesMut would already handle some of the buffer sizing problems, am I wrong?

I don't think BytesMut is involved here since the ser function can be called by anythink with an insuffiently sized buffer.

onur-ozkan: > How would you propose handle someone trying to serialize an invalid bitflag value then ? We could silently slice it of and just mention in the doc that which fields are potentially do that.

We can have `debug_assert!` to cover that. I just don't want to bloat the entire API with dynamic logic which *only* happens when user gives invalid data explicitly.

There's another way with the type system as well. We can create multiple variants of the enum e.g., `GimbalDeviceCapFlags16` and `GimbalDeviceCapFlags32` which can hold the [current contract](https://github.com/mavlink/rust-mavlink/blob/4b45e4d39bc29ebfb6f88f42e625cba3888d7a81/mavlink-bindgen/src/parser.rs#L90-L97) in bindgen but I am not sure how we can make these 2 work on types that use `GimbalDeviceCapFlags`. I may need to think more on this one.

onur-ozkan: > We can create multiple variants of the enum e.g., `GimbalDeviceCapFlags16` and `GimbalDeviceCapFlags32`

Or `GimbalDeviceCapFlags` and `GimbalDeviceCapFlagsExtended`.

onur-ozkan: > I don't think BytesMut is involved here since the ser function can be called by anythink with an insuffiently sized buffer.

Btw, I was confused with the `MavFrame::ser` function. We better rename one of these..

pv42: I considered adding smaller types when required e.g. a GimbalDeviceCapFlagsU16 type that just wraps u16 and implents `Into<GimbalDeviceCapFlags>` and `TryFrom<GimbalDeviceCapFlags>`.

onur-ozkan: Where? I can't see such logic in current PR. I mean if we have specific types for narrow and wider variants then we could avoid dynamic check. e.g., we can make `GimbalDeviceCapFlags` an enum where it takes its variants as a member. We can even avoid the `try_into` part then.

mavlink-bindgen/tests/definitions/oversized_enum.xml
```diff
@@ -0,0 +1,20 @@
...
+  </enums>
+  <messages>
+    <message id="283" name="GIMBAL_DEVICE_INFORMATION">
+      <description>Information about a low level gimbal. This message should be requested by the gimbal manager or a ground station using MAV_CMD_REQUEST_MESSAGE. The maximum angles and rates are the limits by hardware. However, the limits by software used are likely different/smaller and dependent on mode/settings/etc..</description>
+      <field type="uint16_t" name="cap_flags" enum="GIMBAL_DEVICE_CAP_FLAGS">Bitmap of gimbal capability flags.</field>
+      <field type="uint32_t" name="cap_flags2" enum="GIMBAL_DEVICE_CAP_FLAGS" invalid="0">Extended bitmap of gimbal capability flags (32 bit). For backwards compatibility, the lower 16 bits should also be set in cap_flags. Ground stations should prefer this field if non-zero.</field>
+    </message>
+  </messages>
+</mavlink>
+
```
onur-ozkan, line 20: Missing a new line.

mavlink-core/src/lib.rs
pv42: Oh didnt do it I was just considering doing it but ultimatly didn't.
That idea would have the upside that since there is only a single use case of that right now and I don't expect there to be a lot of them ever, most bitflag fields would not change at all.

mavlink-bindgen/tests/definitions/oversized_enum.xml
pv42: ~That is a weird one since there is an empty line at the end but it is still marked as no newline~ nvm the line has a space

mavlink-core/src/lib.rs
onur-ozkan: > That idea would have the upside that since there is only a single use case of that right now and I don't expect there to be a lot of them ever

That's why we should not change the API interface and add a runtime bloat. We can have extended types and solve this problem with the type-system at compile time.

PR #497, opened 2026-05-30
speed up crc computation by migrating to crc-fast

onur-ozkan:
A basic benchmark comparison  I ran on an RP5:

```sh
workload                               frames    crc bytes     passes   crc-any ns/crc  crc-fast ns/crc    speedup

command_long v2                             1           39    3441480           156.41            74.53      2.10x
heartbeat v2                               46          828     162098            82.98            72.57      1.14x
servo_output_raw v2                        37         1702      78858           181.34            75.82      2.39x
raw_imu v2                                 37         1406      95460           153.21            75.84      2.02x
gps_raw_int v2                             37         2257      59467           234.13            77.18      3.03x
system_time v2                             36          756     177536            93.56            72.42      1.29x
```

So we get a good amount of speed simply by replacing the dependency.

Note that `crc-fast` is not only a faster CRC implementation but it is also more widely used. According to crates.io, as of today, it has over 18 million more downloads than `crc-any`.

Blocked by: https://github.com/awesomized/crc-fast-rust/pull/49

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ byteorder = { version = "1.3.4", default-features = false }
 
 [workspace.package]
 edition = "2024"
-rust-version = "1.85.0"
+rust-version = "1.89.0"
 keywords = ["mavlink", "parser", "protocol", "embedded"]
 categories = ["aerospace", "aerospace::protocols", "parsing", "embedded"]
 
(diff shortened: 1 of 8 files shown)


onur-ozkan:
I want to measure this on one of my ARM machines but I don't have access to any of them at the moment.

Until then, I will keep this as a draft.

PR #498, opened 2026-05-30
udp: avoid per-send allocation for outgoing frames

onur-ozkan:
Avoids per-send allocation and copy for udp send.

### DIFF
--- a/mavlink-core/src/connection/udp/async.rs
+++ b/mavlink-core/src/connection/udp/async.rs
@@ -7,7 +7,7 @@ use std::{collections::VecDeque, io::Read, sync::Arc};
 use async_trait::async_trait;
 use futures::lock::Mutex;
 use tokio::{
-    io::{AsyncRead, ReadBuf},
+    io::{AsyncRead, AsyncWrite, ReadBuf},
     net::UdpSocket,
 };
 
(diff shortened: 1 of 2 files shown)


pv42:
In my testing this seems slightly slower, but it probaby heavily depends on the average payload lenght of the messages serialized. Maybe using `Vec::with_capacity(~10)` is optimal, probably also dependent on OS, CPU architechture, cache size etc.



Bench I used:

test benchs::bench_udp_stack ... bench:      12,668.58 ns/iter (+/- 170.58)
test benchs::bench_udp_vec   ... bench:      12,642.90 ns/iter (+/- 118.51)

```rs
    #[bench]
    pub fn bench_udp_stack(bencher: &mut Bencher) {
        use mavlink::dialects::common::MavMessage;
        use rand::{SeedableRng, rngs::StdRng};
        use mavlink_core::{MAX_FRAME_SIZE, Message};
        const COUNT: usize = 255;
        let mut rng = StdRng::seed_from_u64(42);

        let mut data: Vec<MavMessage> = Vec::with_capacity(COUNT);
        for id in 0..COUNT {
            match MavMessage::random_message_from_id(id as u32, &mut rng) {
                Some(msg) => data.push(msg),
                None => (),
            }
        }
        bencher.iter(|| {
            for msg in &data {
                let mut buf = [0u8; MAX_FRAME_SIZE];
                let mut cursor = &mut buf[..];
                
                let header = mavlink::MavHeader { system_id: 0, component_id: 0, sequence: 0 };
                mavlink::write_versioned_msg(&mut cursor, MavlinkVersion::V2, header, msg).unwrap();
            }
        });
    }
    #[bench]
    pub fn bench_udp_vec(bencher: &mut Bencher) {
        // same setup [..]
        bencher.iter(|| {
            for msg in &data {
                let mut buf = Vec::new();
                
                let header = mavlink::MavHeader { system_id: 0, component_id: 0, sequence: 0 };
                mavlink::write_versioned_msg(&mut buf, MavlinkVersion::V2, header, msg).unwrap();
            }
        });
    }
        
```

onur-ozkan:
> In my testing this seems slightly slower, but it probaby heavily depends on the average payload lenght of the messages serialized. Maybe using `Vec::with_capacity(~10)` is optimal, probably also dependent on OS, CPU architechture, cache size etc.
> 
> Bench I used:
> 
> test benchs::bench_udp_stack ... bench: 12,668.58 ns/iter (+/- 170.58) test benchs::bench_udp_vec ... bench: 12,642.90 ns/iter (+/- 118.51)
> 
> ```rust
>     #[bench]
>     pub fn bench_udp_stack(bencher: &mut Bencher) {
>         use mavlink::dialects::common::MavMessage;
>         use rand::{SeedableRng, rngs::StdRng};
>         use mavlink_core::{MAX_FRAME_SIZE, Message};
>         const COUNT: usize = 255;
>         let mut rng = StdRng::seed_from_u64(42);
> 
>         let mut data: Vec<MavMessage> = Vec::with_capacity(COUNT);
>         for id in 0..COUNT {
>             match MavMessage::random_message_from_id(id as u32, &mut rng) {
>                 Some(msg) => data.push(msg),
>                 None => (),
>             }
>         }
>         bencher.iter(|| {
>             for msg in &data {
>                 let mut buf = [0u8; MAX_FRAME_SIZE];
>                 let mut cursor = &mut buf[..];
>                 
>                 let header = mavlink::MavHeader { system_id: 0, component_id: 0, sequence: 0 };
>                 mavlink::write_versioned_msg(&mut cursor, MavlinkVersion::V2, header, msg).unwrap();
>             }
>         });
>     }
>     #[bench]
>     pub fn bench_udp_vec(bencher: &mut Bencher) {
>         // same setup [..]
>         bencher.iter(|| {
>             for msg in &data {
>                 let mut buf = Vec::new();
>                 
>                 let header = mavlink::MavHeader { system_id: 0, component_id: 0, sequence: 0 };
>                 mavlink::write_versioned_msg(&mut buf, MavlinkVersion::V2, header, msg).unwrap();
>             }
>         });
>     }
>         
> ```

You have pre-allocation on the `Vec`, current tree doesn't do that. So it's not the same thing. I will check (later) if pre-allocated `Vec` is beatable on stack.

onur-ozkan:
Now we don't perform any allocations or copies so it should be faster than before.

As for the benchmark you shared the slowdown was caused by the following factors:

- ~Pre-allocating the `Vec` (which the original implementation doesn't do).~ I was looking on the wrong `Vec`.
- The message size. The stack allocation used the full message size which is just a waste for small messages but should be faster on large messages.

The current approach should outperform both.

PR #499, opened 2026-05-30
refactor: make `MavFrame::ser` no-copy

onur-ozkan:
Writes payload straight into the destination buffer right after the header and avoids the intermediate buffer and an extra copy.

Speeds up the `MavFrame::ser` around 20% on a regular x86 CPU.


pv42:
Looks like a straight up improvement, `put_slice` becomes unused with this and could potentially removed.

onur-ozkan:
It is still used.

PR #503, opened 2026-06-02
allow rustdoc::broken_intra_doc_links lint in generated code

pv42:
Supresses a lot of unresolved link warning on `cargo doc` that trigger on array indices/general square brackets from mavlinks descriptions (e.g. "data[1]").


onur-ozkan (approved):
Looks right

PR #504, opened 2026-06-03
Update rand to 0.10, remove unused bindgen dependencies

pv42:
- update to rand [0.10.0](https://github.com/rust-random/rand/releases/tag/0.10.1) and [0.10.1](https://github.com/rust-random/rand/releases/tag/0.10.1)
- renamed `rand::RngCore` to `rand::Rng`
- the `Rng` trait does no longer implement `DerefMut`, but `TryRng<Error = std::convert::Infallible>` does implement both `DerefMut` and `Rng` therefore the change to the `Message` trait.
- the rand and arbitrary deps were not needed in bindgen.

### DIFF
--- a/mavlink-bindgen/Cargo.toml
+++ b/mavlink-bindgen/Cargo.toml
@@ -16,7 +16,6 @@ categories.workspace = true
 anstyle = { version = "1.0.11", optional = true }
 anstyle-parse = { version = "1.0.0", optional = true }
 anstyle-query = { version = "1.1.4", optional = true }
-arbitrary = { version = "1.4", optional = true, features = ["derive"] }
 clap = { version = "~4.6.0", optional = true, default-features =false, features = ["derive", "help", "usage", "error-context"] }
 clap_builder = { version = "4.3.24", optional = true}
 clap_lex = { version = "1.0.0", optional = true }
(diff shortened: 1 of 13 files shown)


mavlink-bindgen/Cargo.toml
```diff
@@ -16,15 +16,13 @@ categories.workspace = true
 anstyle = { version = "1.0.11", optional = true }
 anstyle-parse = { version = "1.0.0", optional = true }
 anstyle-query = { version = "1.1.4", optional = true }
-arbitrary = { version = "1.4", optional = true, features = ["derive"] }
```
onur-ozkan, line 19: > the rand and arbitrary deps were not needed in bindgen.

`arbitrary::Arbitrary` is used in bindgen and your PR literally touched where `rand` is also used.

pv42: Both are only used inside quote! macros and those don't need the dependencies to be present to compile.

onur-ozkan: Cool, can you add comments on them (if it's possible without injecting those comments to the generated code) so we don't get confused later again?

pv42: Done, also seperated the changes into 2 commits and rebased.

onur-ozkan (approved):

PR #507, opened 2026-06-03
Add non-zero timeout for serial connection

akumaigorodski:
With default zero timeout it was taking 100% of assigned CPU core on RPI compute module 5.


mavlink-core/src/connection/direct_serial/sync.rs
```diff
@@ -105,6 +106,7 @@ impl Connectable for SerialConfig {
             .parity(Parity::None)
             .stop_bits(StopBits::One)
             .flow_control(FlowControl::None)
+            .timeout(Duration::from_millis(1))
```
onur-ozkan, line 109: This should rather be a configurable value not a hard-coded one so you can set it however you would like.

onur-ozkan (changes requested):

PR #509, opened 2026-06-06
add timeout configuration for sync serial reads

onur-ozkan:
Adds timeout configuration for serial reads.

### DIFF
--- a/mavlink-core/src/connection/direct_serial/config.rs
+++ b/mavlink-core/src/connection/direct_serial/config.rs
@@ -1,4 +1,5 @@
 use core::fmt::Display;
+use core::time::Duration;
 
 /// MAVLink address for a serial connection
 ///
@@ -14,6 +15,7 @@ use core::fmt::Display;
 pub struct SerialConfig {
     pub(crate) port_name: String,
     pub(crate) baud_rate: u32,
+    pub(crate) timeout: Option<Duration>,
     read_buffer_capacity: usize,
 }
 
@@ -26,8 +28,21 @@ impl SerialConfig {
         Self {
             port_name,
             baud_rate,
+            timeout: None,
             read_buffer_capacity: default_capacity,
         }
+    }
+
+    /// Sets the serial port timeout.
+    ///
+    /// When set, serial reads will return an error after the specified duration.
+    ///
+    /// By default, this is 1 millisecond.
+    ///
+    /// Only applies to sync connections and is a no-op for async ones.
+    pub fn timeout(mut self, timeout: Duration) -> Self {
+        self.timeout = Some(timeout);
+        self
     }
 
     /// Updates the read buffer capacity.
--- a/mavlink-core/src/connection/direct_serial/sync.rs
+++ b/mavlink-core/src/connection/direct_serial/sync.rs
@@ -106,7 +106,7 @@ impl Connectable for SerialConfig {
             .parity(Parity::None)
             .stop_bits(StopBits::One)
             .flow_control(FlowControl::None)
-            .timeout(Duration::from_millis(1))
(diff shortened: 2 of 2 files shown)


onur-ozkan:
Pinging @akumaigorodski so they know their change is updated.

patrickelectric:
Just to be sure, what was broken ?

onur-ozkan:
> Just to be sure, what was broken ?

recv() can time out after 1 millisecond when there are no messages available. If someone wants non-blocking, one-shot read they can use try_read instead. A 1 millisecond timeout will be as instant as try_recv() therefore makes no sense IMO.

akumaigorodski:
Factually my original change fixed 100% CPU core issue while retaining all the functionality. My interpretation of that is without idle timeout it was retrying to read immediately on failure, thus taking all available compute resources. 

And then staying idle for 1ms seemed like good enough tradeoff between possible added latency and compute use.

Just in case, 100% CPU was not manifesting itself when RPI was connected to SITL simulator via Ethernet/UDP, but started showing up with UART and actual flight controller.

I don't quite like the way it is implemented now, I'd rather remove factory method with hidden logic entirely and let users, for example, provide their own `serialport::new` (because someone might need different stop bits etc), adding utility methods for every parameter is not scalable.

onur-ozkan:
I didn't understand what you mean with hidden logic and scalibility to be honest. Can you say more? Scale what exactly? What is hidden ?

akumaigorodski:
Basically I'm proposing to remove a higher level `mavlink::connect` with different types of strings and its parsing logic, and let users assemble each invariant from their own parts. If this was in place then I would never have an issue configuring my own `serialport::new`.

onur-ozkan:
> > Just to be sure, what was broken ?
> 
> recv() can time out after 1 millisecond when there are no messages available. If someone wants non-blocking, one-shot read they can use try_read instead. A 1 millisecond timeout will be as instant as try_recv() therefore makes no sense IMO.

My claim about recv() should be wrong here. I forgot that we are doing a loop inside recv(). I thought the timeout was applied to whole recv().

So there shouldn't be anything broken really.

onur-ozkan:
> Basically I'm proposing to remove a higher level `mavlink::connect` with different types of strings and its parsing logic, and let users assemble each invariant from their own parts. If this was in place then I would never have an issue configuring my own `serialport::new`.

You could use try_recv to avoid all this. Then you can have your own way to receive messages with any specific timeout.

onur-ozkan:
Changed the approach. The default timeout now remains at 1 ms. I think the 100% CPU usage was caused by having no timeout i.e the lopp was constantly spinning in recv(). With a 1 ms timeout each iteration is blocked by the serial port for up to 1 ms which prevents the busy loop. From that aspect the default 1ms timeout makes sense. But it's still worth keeping it configurable since different use cases may benefit from different timeout values.

mprudnik-sf:
@onur-ozkan Any update on this? We hit the same missing-timeout issue and this would help us a lot. Happy to test if useful.

onur-ozkan:
> @onur-ozkan Any update on this? We hit the same missing-timeout issue and this would help us a lot. Happy to test if useful.

cc @patrickelectric

patrickelectric (approved):

mprudnik-sf:
@patrickelectric do you plan a release soon? This is already in master but not on crates.io yet. Would highly appreciate it. Thank you!

onur-ozkan:
> @patrickelectric do you plan a release soon? This is already in master but not on crates.io yet. Would highly appreciate it. Thank you!

There's a lot of new stuff in current master branch and I am holding the release for some of the current open PRs. It's already going to be a major release so it's better to include other big changes as well. Until then you can use the git commit hash from master as a version instead of pulling directly from crates.io.

PR #511, opened 2026-06-22
test dialects dynamically

### LINKED ISSUE #455: make sure we don't miss any dialect
We could have a CI job that ensures we don't miss any dialect from `mavlink/message_definitions`.

onur-ozkan:
`scripts/list_dialect_features.py` prints all dialect features resolved directly from the submodule and CI uses them to test dialects. If any dialect is missing, the test build will fail with:

    error: none of the selected packages contains this feature: $feature

Fixes https://github.com/mavlink/rust-mavlink/issues/455

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,12 +24,22 @@ jobs:
         with:
           args: --all --all-targets --features format-generated-code --features mav2-message-signing --features tokio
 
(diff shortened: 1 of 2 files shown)


onur-ozkan:
CI is not running :thinking:

joaoantoniocardoso (approved):

PR #518, opened 2026-07-03
create stable api for mavlink frame constants

onur-ozkan:
Expose MAVLink frame layout constants through a new public stable API module `consts` with shared values at the module root and version-specific values under `consts::v1` and `consts::v2`.

Also clean up the in-tree implementation to use named values instead of hard-coding frame layout numbers.

### DIFF
--- a/mavlink-core/src/consts.rs
+++ b/mavlink-core/src/consts.rs
@@ -1,0 +1,52 @@
+//! MAVLink frame constants.
+
+/// Size of the STX marker.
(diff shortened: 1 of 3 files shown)


joaoantoniocardoso (approved):

0rlych1kk4:
This is a useful API improvement. The shared and version-specific constant structure should make downstream frame handling clearer and less dependent on hard-coded layout values. I’ll check whether existing examples can be updated to demonstrate the new stable API.

PR #502, opened 2026-06-02
disable ts-rs feature for docs.rs

pv42:
Helps #501, `ts-rs` does not generate any docs anyways.

Without the feature the RAM usage is ~2GB less:
without `ts-rs`
```
cargo rustdoc -p mavlink --features serde,format-generated-code,arbitrary,tokio,embedded,std,dialect-all,mav2-message-extensions,mav2-message-signing,dialect-all,dialect-ardupilotmega,dialect-asluav,dialect-avssuas,dialect-common,dialect-csairlink,dialect-cubepilot,dialect-development,dialect-icarous,dialect-loweheiser,dialect-marsh,dialect-matrixpilot,dialect-minimal,dialect-paparazzi,dialect-python_array_test,dialect-standard,dialect-stemstudios,dialect-storm32,dialect-test,dialect-ualberta,dialect-uavionix  -j 1 -- -Z time-passes
 Documenting mavlink v0.18.0 (E:\pv42\rust-mavlink\mavlink)
time:   0.000; rss:   20MB ->   20MB (   +0MB)  parse_crate                                                                                                                                                                                 
time:   0.000; rss:   23MB ->   23MB (   +0MB)  crate_injection                                                                                                                                                                             
time:  29.987; rss:   23MB -> 5841MB (+5817MB)  expand_crate                                                                                                                                                                                
time:  29.987; rss:   23MB -> 5841MB (+5817MB)  macro_expand_crate
time:   0.620; rss: 5841MB -> 5841MB (   +1MB)  AST_validation                                                                                                                                                                              
time:   0.102; rss: 5841MB -> 6033MB ( +192MB)  finalize_imports                                                                                                                                                                            
time:   0.021; rss: 6033MB -> 6039MB (   +6MB)  compute_effective_visibilities                                                                                                                                                              
time:   0.062; rss: 6039MB -> 6039MB (   +0MB)  lint_reexports                                                                                                                                                                              
time:   0.213; rss: 6039MB -> 6011MB (  -28MB)  finalize_macro_resolutions                                                                                                                                                                  
time:   5.522; rss: 6011MB -> 6801MB ( +789MB)  late_resolve_crate                                                                                                                                                                          
time:   0.503; rss: 6801MB -> 6803MB (   +2MB)  resolve_check_unused                                                                                                                                                                        
time:   0.874; rss: 6803MB -> 6803MB (   +0MB)  resolve_postprocess                                                                                                                                                                         
[...]

PR #519, opened 2026-07-09
fix: reduce docs.rs build features for mavlink

0rlych1kk4:
This PR reduces the docs.rs build feature set for the mavlink crate.

The docs.rs build for mavlink 0.18.0 is currently failing. The failed docs.rs build took 14m 24s, which is close to the docs.rs rustdoc execution limit of 15m. The previous docs.rs metadata enabled many generated dialect modules and optional integrations, making the documentation build expensive.

This change configures docs.rs to build with a smaller explicit feature set:

- std
- dialect-common
- format-generated-code

This should reduce docs.rs build time while still producing useful public crate documentation.

Tested locally with:

RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc -p mavlink --no-deps --no-default-features --features "std,dialect-common,format-generated-code"

Result:

Finished in 8.13s

Also tested:

cargo test --workspace

Result:

All workspace tests passed.

### DIFF
--- a/mavlink/Cargo.toml
+++ b/mavlink/Cargo.toml
@@ -1,4 +1,3 @@
-
 [package]
 name = "mavlink"
 version = "0.18.0"
@@ -100,8 +99,8 @@ format-generated-code = []
 ts-rs = ["dep:ts-rs"]
 
 [package.metadata.docs.rs]
-# Build with all features except ts-rs on docs.rs so that users viewing documentation
-# can see everything
+# Build docs.rs with the default feature set and a reduced set of dialects to avoid
+# hitting docs.rs sandbox resource limits.
 features = [
     "default",
     "mav2-message-extensions",
@@ -110,28 +109,12 @@ features = [
     "tokio",
     "arbitrary",
 
-    "dialect-all",
     "dialect-ardupilotmega",
-    "dialect-asluav",
-    "dialect-avssuas",
     "dialect-common",
-    "dialect-csairlink",
-    "dialect-cubepilot",
-    "dialect-development",
-    "dialect-icarous",
-    "dialect-loweheiser",
-    "dialect-marsh",
-    "dialect-matrixpilot",
     "dialect-minimal",
-    "dialect-paparazzi",
-    "dialect-python_array_test",
     "dialect-standard",
-    "dialect-stemstudios",
-    "dialect-storm32",
-    "dialect-test",
-    "dialect-ualberta",
-    "dialect-uavionix",
 ]
+rustdoc-args = ["--cfg", "docsrs"]
 
 [dev-dependencies]
 tokio = { version = "1.0", default-features = false, features = ["macros", "rt", "time" ] }


pv42:
The bottleneck is usually memory (https://docs.rs/crate/mavlink/0.18.0/builds/3255271), and I belive removing `ts-rs` (#502) already fixed that somewhat. If we want to have a bigger margin of free ram so it doesn't start failing again we can remove more dialect features. 
This PR removes virtually all features (including default ones) which seems excessive.

patrickelectric:
We can just request more resources in the sandbox (as done in the pass). Or just improve the compilation itself and find the bottleneck.

0rlych1kk4:
Thanks @pv42 for the feedback. Agreed, the first version reduced the docs.rs feature set too aggressively.
I’ve updated the PR to keep the default and main optional features, and only reduce the dialect list to give docs.rs more resource margin.

The updated docs.rs feature set keeps:
- default
- mav2-message-extensions
- mav2-message-signing
- embedded
- tokio
- arbitrary
- dialect-ardupilotmega
- dialect-common
- dialect-minimal
- dialect-standard

Local validation:

RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc -p mavlink --no-deps --features "default,mav2-message-extensions,mav2-message-signing,embedded,tokio,arbitrary,dialect-ardupilotmega,dialect-common,dialect-minimal,dialect-standard"

Finished in 18.78s.

cargo test --workspace also passed.

patrickelectric:
I would vote against removing the dialects since they are really useful for development in the docs.

0rlych1kk4:
Understood, thanks @patrickelectric.
I’ll close this PR then, and I’ll take a separate look at the actual build bottleneck instead and see if there is a smaller improvement that keeps the docs feature coverage intact.

0rlych1kk4:
Closing this PR as discussed.

0rlych1kk4:
Hi @patrickelectric and @pv42 I did a bit more local investigation after closing this.
A clean full docs.rs-style build completed locally in about 5m59s, but the generated output is quite large:

- target/doc/mavlink: 497M
- target/doc/mavlink files: 6,889
- generated mavlink out dir: 29M
- generated files: 22
- generated Rust lines: 660,270

Largest generated files:

- all.rs: 63,152 lines
- storm32.rs: 54,753 lines
- ardupilotmega.rs: 52,837 lines
- matrixpilot.rs: 42,852 lines
- asluav.rs: 41,715 lines
- development.rs: 41,572 lines
- uavionix.rs: 40,548 lines
- common.rs: 39,278 lines

Some local timing comparisons:

- default docs build: 16.69s
- optional features only: 13.74s
- all dialects only: 1m39s
- full docs.rs-style feature set, clean build: 5m59s

So I agree that removing dialect coverage is not the right direction. The pressure appears to come mainly from documenting the generated dialect modules, especially when all dialects are combined with the full optional feature set.

I’ll continue looking for a possible improvement that keeps dialect coverage intact.

pv42:
As mentioned before the docs.rs bottleneck is usually ram usage (improvents in other areas are obviously also welcome). You can see the memory usage with `cargo rustdoc -- -Z time-passes`.

PR #523, opened 2026-07-22
feat: runtime loadable custom dialect support

onur-ozkan:
rust-mavlink currently supports only compile-time generated dialects. Users with a vehicle-specific or non-standard dialect must add the XML definition, regenerate bindings and rebuild their application before they can communicate with it.

That is not practical for applications that need to support arbitrary vehicles or allow users to provide their own dialect definitions. This change adds runtime-loaded dialect support so an application can accept a dialect XML path or uploaded file and communicate with that dialect without code generation and rebuilds.

Changes are simply:

- Add the `Dialect` trait and dialect-aware read/write helpers for sync and async transports.
- Keep generated dialect support unchanged through a static dialect adapter.
- Add `DynamicDialect` behind the `dialect-dynamic` feature.
- Add dialect-aware connection helpers including `connect_with_dialect` and async equivalents.
- Add a `runtime-dialect` serial example and usage documentation.

1:1 byte representation is covered with a unit test in `dynamic` module.


pv42:
I read through some parts of it and here are some early comments: 
- consider renaming the Mavtype::len() to size() if it becomes public to be in line with other rust apis (would fix the lint)
- returning InvalidPayloadLength on payloads to long is problematic since the sender could use an mavlink 2 extension that the current dialect is not yet aware of

onur-ozkan:
Do we have any bandwith to review this PR as a release candidate?

Issue #486, opened 2026-04-05
Add profiling support for performance analysis


onur-ozkan:
It would be useful to add profiling support to rust-mavlink so we can analyze performance characteristics rather than guessing the runtime impact when reviewing PRs or making changes to the project's core.

onur-ozkan:
It's not clear yet which approach to use. As far as I can tell `perf` is one of the necessary/key tools we should utilize in this flow.

0rlych1kk4:
@onur-ozkan  I would like to work on this. I can start by adding a small, reproducible profiling setup around the main parsing and serialization paths, using perf on Linux and documenting how to generate flamegraphs.
Before I begin, would you prefer the profiling workload to use existing examples/tests, or should I add a dedicated benchmark-style binary with representative MAVLink v1 and v2 messages? If this direction looks appropriate, could you assign the issue to me?

0rlych1kk4:
@onur-ozkan  Just following up on this. I’m still interested in working on the profiling support and can begin with a small Linux perf-based setup using representative MAVLink v1 and v2 workloads. Please let me know whether you prefer this to use the existing examples or a dedicated profiling binary.

onur-ozkan:
I have no architecture on mind so if you have any, please share so we can discuss on that.

0rlych1kk4:
@onur-ozkan  Thanks. I suggest starting with a minimal dedicated profiling binary rather than coupling profiling to the existing examples or test suite.
My initial proposal is:

- deterministic MAVLink v1 and v2 workloads
- separate parsing and serialization loops
- representative small and larger messages
- release builds with debug symbols for useful perf output
- documented perf record, perf report, and flamegraph steps
- no public API changes or additional runtime dependencies

I would start with this limited scope and open a draft PR so we can review the structure before expanding it. Does this direction sound reasonable? If so, could you also assign the issue to me?

0rlych1kk4:
@onur-ozkan  I have opened draft PR #528 with the initial profiling implementation based on the approach discussed here. It currently keeps the scope limited to deterministic MAVLink v1/v2 parsing and serialization workloads, with no public API changes. Happy to adjust the structure based on your review.

onur-ozkan:
Thanks, I will try to review it during the week.

0rlych1kk4:
@onur-ozkan  Quick update on #528 : I’ve reworked the draft based on your review and removed the earlier manual profiling workload.
[...]

PR #528, opened 2026-08-09
feat: add profiling support for MAVLink performance analysis

0rlych1kk4:
Refs #486 

## Summary

Adds an initial CI-based performance measurement workflow for rust-mavlink.

For each pull request, the workflow compares the PR head against the base
revision using the same release build configuration and reports:

- compilation time
- `libmavlink` artifact size
- percentage change between base and PR

The results are published in the GitHub Actions job summary.

## Implementation

- adds .github/workflows/performance.yml
- adds scripts/measure_build.sh
- measures a clean release build using dialect-common
- compares the base revision with the PR revision
- reports compilation time and artifact size
- publishes results via $GITHUB_STEP_SUMMARY, so PR write permissions are not required

## Current scope

This first iteration intentionally focuses on the two metrics suggested in #486:

- compilation time
- build artifact size

Runtime performance, resource usage, visualizations, and automatic PR comments
can be considered as follow-up work.

## Validation

The Performance workflow completed successfully and produced a base-vs-PR comparison in the Actions summary.
The existing Test all targets workflow also completed successfully.
This remains a draft while the CI approach is reviewed.


onur-ozkan (commented):
Hmm. This is quite far from what I had in mind. You basically wrote a sample program to be measured with the `perf` command, entirely manually. The main motivation is CI, let's not forget that.

Ideally, we should have a script that can comment on each PR with information about how it affects library compilation time, binary size, performance, resource usage etc., all likely with some visualizations if possible.

As a first step though, we can simply check the compilation time and build size. We can improve it with follow-up work later.

0rlych1kk4:
@onur-ozkan  Thanks, that clarifies the goal. I focused too much on providing a manual perf workload and missed that the main value should come from CI feedback on PRs.
I’ll rework the initial scope around the first step you suggested: measuring compilation time and build size in CI. My plan is to keep it small initially:

- build a defined rust-mavlink configuration in release mode;
- capture compilation duration;
- capture resulting artifact size;
- expose the results in the CI job output;
- structure it so PR comments/comparisons against the base branch can be added next

I’ll update the draft PR accordingly and keep runtime profiling as possible follow-up work.

0rlych1kk4:
@onur-ozkan  I’ve reworked #528 around the CI-first direction you described and removed the earlier manual profiling workload. The draft now focuses on comparing PR vs base compilation time and libmavlink artifact size in CI, with the result rendered in the Actions summary. The Performance workflow and the existing full test workflow both pass. Happy to adjust the measurement approach or build configuration based on your review.

.github/workflows/performance.yml
```diff
@@ -0,0 +1,91 @@
...
+          echo "artifact_size_bytes=$artifact_size_bytes" >> "$GITHUB_OUTPUT"
+
+      - name: Generate comparison
+        env:
+          BASE_TIME: ${{ steps.base.outputs.compile_time_ms }}
+          PR_TIME: ${{ steps.pr.outputs.compile_time_ms }}
+          BASE_SIZE: ${{ steps.base.outputs.artifact_size_bytes }}
+          PR_SIZE: ${{ steps.pr.outputs.artifact_size_bytes }}
+        run: |
+          python3 - <<'PY' >> "$GITHUB_STEP_SUMMARY"
```
onur-ozkan, line 59: Let's not write python programs inside a CI flow. We should have proper tooling scripts inside scripts directory (this helps us to reproduce CI steps in our machines as well) and use them from CI.

0rlych1kk4: Hi @onur-ozkan  Agreed. I’ll move the comparison/report generation out of the workflow and into a reusable script under scripts/, so the same logic can be run locally as well as from CI.

0rlych1kk4: @onur-ozkan updated in the latest commit. The comparison/report generation now lives in scripts/compare_build_metrics.py, and the workflow invokes the reusable script. I also verified that the script can be run locally.

PR #529, opened 2026-08-19
Added socket-backed UDP and TCP configs for caller-provided sockets

joaofrf:
Allow UDP and TCP configurations to use outside sockets.

Added:
- UdpConfig::from_socket
- TcpConfig::from_stream
- TcpConfig::from_listener

Note:
I had previously started this and finally had the time to finish:
Original pr: https://github.com/mavlink/rust-mavlink/pull/438

### DIFF
--- a/mavlink-core/src/connection/tcp/config.rs
+++ b/mavlink-core/src/connection/tcp/config.rs
@@ -1,4 +1,7 @@
 use core::fmt::Display;
+use std::io;
+use std::net::{TcpListener, TcpStream};
+use std::sync::{Arc, Mutex};
 
 /// Type of TCP connection
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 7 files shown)


patrickelectric (approved):

mavlink-core/src/connection/tcp/config.rs
```diff
@@ -23,12 +26,80 @@ pub enum TcpMode {
...
+    /// Creates a TCP client configuration from an already connected stream.
+    ///
+    /// Socket-backed configurations are one-shot, including across clones.
+    pub fn from_stream(stream: TcpStream) -> io::Result<Self> {
+        let address = stream.peer_addr()?.to_string();
+        Ok(Self {
+            address,
+            mode: TcpMode::TcpOut,
+            source: TcpSource::Stream(Arc::new(Mutex::new(Some(stream)))),
+        })
```
patrickelectric, line 58: just to be sure... this should have a timeout as well, right ?

joaofrf: Hey Patrick, sorry i missed your comment.
the caller can set it before passing the stream:
```
stream.set_read_timeout(Some(Duration::from_millis(100)))?;

let config = TcpConfig::from_stream(stream)?;
```

i believe 100ms is the default, but i'd have to double check

PR #530, opened 2026-08-19
Added a new method to recv including unknown messages

joaofrf:
New method that allows receiving unknown mavlink messages unknown to our mavlink definition source.

This is something I've had for a while and decided to PR since this is very useful in cases where we're dealing with custom mavlink definitions and one of the softwares is still missing the messages because it wasn't updated on time

### DIFF
--- a/mavlink-core/src/connection/async.rs
+++ b/mavlink-core/src/connection/async.rs
@@ -24,6 +24,14 @@ pub trait AsyncMavConnection<M: Message + Sync + Send> {
     /// Yield until a valid frame is received, ignoring invalid messages.
     async fn recv_raw(&self) -> Result<MAVLinkMessageRaw, MessageReadError>;
 
+    /// Receive a raw MAVLink message, including messages whose ID is unknown to `M`.
+    ///
+    /// Unknown messages are returned without CRC validation because their
+    /// dialect-specific `CRC_EXTRA` is unavailable.
+    async fn recv_raw_including_unknown(&self) -> Result<MAVLinkMessageRaw, MessageReadError> {
+        self.recv_raw().await
+    }
+
     /// Try to receive a MAVLink message.
     ///
     /// Non-blocking variant of `recv()`, returns immediately with a `MessageReadError`
--- a/mavlink-core/src/connection/direct_serial/async.rs
+++ b/mavlink-core/src/connection/direct_serial/async.rs
@@ -14,7 +14,7 @@ use crate::connection::AsyncConnectable;
 use crate::connection::direct_serial::config::SerialConfig;
 use crate::connection_shared::{
     ConnectionState, next_atomic_send_header, read_message_async, read_raw_message_async,
-    write_message_async, write_raw_message_async,
+    read_raw_message_async_including_unknown, write_message_async, write_raw_message_async,
 };
 use crate::error::MessageReadError;
 use crate::{MavHeader, MavlinkVersion, Message, async_peek_reader::AsyncPeekReader};
@@ -53,6 +53,24 @@ impl<M: Message + Sync + Send> AsyncMavConnection<M> for AsyncSerialConnection {
             let result = read_raw_message_async::<M, _>(port.deref_mut(), &self.state).await;
             match result {
                 Ok(message) => return Ok(message),
+                Err(MessageReadError::Io(e)) if e.kind() == io::ErrorKind::UnexpectedEof => {
+                    return Err(MessageReadError::Io(e));
+                }
+                _ => {}
+            }
+        }
+    }
+
+    async fn recv_raw_including_unknown(
+        &self,
+    ) -> Result<MAVLinkMessageRaw, crate::error::MessageReadError> {
+        let mut port = self.read_port.lock().await;
+        loop {
+            let result =
+                read_raw_message_async_including_unknown::<M, _>(port.deref_mut(), &self.state)
+                    .await;
+            match result {
+                ok @ Ok(..) => return ok,
                 Err(MessageReadError::Io(e)) if e.kind() == io::ErrorKind::UnexpectedEof => {
                     return Err(MessageReadError::Io(e));
                 }
--- a/mavlink-core/src/connection/direct_serial/sync.rs
+++ b/mavlink-core/src/connection/direct_serial/sync.rs
@@ -3,8 +3,8 @@
 use crate::Connectable;
 use crate::connection::{Connection, MavConnection};
 use crate::connection_shared::{
-    ConnectionState, next_atomic_send_header, read_message, read_raw_message, write_message,
-    write_raw_message,
+    ConnectionState, next_atomic_send_header, read_message, read_raw_message,
+    read_raw_message_including_unknown, write_message, write_raw_message,
 };
 use crate::error::{MessageReadError, MessageWriteError};
 use crate::peek_reader::PeekReader;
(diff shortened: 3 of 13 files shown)


onur-ozkan (commented):
How do you know if it's corrupted packet or valid unknown packet?

Ideally one should simply use https://github.com/mavlink/rust-mavlink/pull/523 or compile it with mavlink-bidgen if they have modified dialect.

joaofrf:
@onur-ozkan 
> How do you know if it's corrupted packet or valid unknown packet?
> 
> Ideally one should simply use #523 or compile it with mavlink-bidgen if they have modified dialect.

We can’t really distinguish them since we dont have crc_extra
In my opinion it is fine to forward it anyways since whoever really needs to read that packet data will have to parse it

I can rename methods to something like "recv_raw_including_unvalidated" if that's prefered

onur-ozkan:
> @onur-ozkan
> 
> > How do you know if it's corrupted packet or valid unknown packet?
> > Ideally one should simply use #523 or compile it with mavlink-bidgen if they have modified dialect.
> 
> We can’t really distinguish them since we dont have crc_extra In my opinion it is fine to forward it anyways since whoever really needs to read that packet data will have to parse it
> 

I disagree, sorry. If you expect different messages there are multiple ways to receive them (like I mentioned above). We cannot expose an insecure API that tries to resolve random payload.

joaofrf:
> > @onur-ozkan
> > > How do you know if it's corrupted packet or valid unknown packet?
> > > Ideally one should simply use #523 or compile it with mavlink-bidgen if they have modified dialect.
> > 
> > 
> > We can’t really distinguish them since we dont have crc_extra In my opinion it is fine to forward it anyways since whoever really needs to read that packet data will have to parse it
> 
> I disagree, sorry. If you expect different messages there are multiple ways to receive them (like I mentioned above). We cannot expose an insecure API that tries to resolve random payload.

How would you go about this?
I understand your preference for #523 but in some cases, where I use this library, and I really use it a lot, It wouldn't be feasible.
I understand if you decide not to merge this or guide it in that direction, I'm just honestly interested in this feature :)

onur-ozkan:
> > > @onur-ozkan
> > > > How do you know if it's corrupted packet or valid unknown packet?
> > > > Ideally one should simply use #523 or compile it with mavlink-bidgen if they have modified dialect.
> > > 
> > > 
> > > We can’t really distinguish them since we dont have crc_extra In my opinion it is fine to forward it anyways since whoever really needs to read that packet data will have to parse it
> > 
> > 
> > I disagree, sorry. If you expect different messages there are multiple ways to receive them (like I mentioned above). We cannot expose an insecure API that tries to resolve random payload.
> 
> How would you go about this? I understand your preference for #523 but in some cases, where I use this library, and I really use it a lot, It wouldn't be feasible. I understand if you decide not to merge this or guide it in that direction, I'm just honestly interested in this feature :)

What is your use case? Can you explain more?

joaofrf:
> > > > @onur-ozkan
> > > > > How do you know if it's corrupted packet or valid unknown packet?
> > > > > Ideally one should simply use #523 or compile it with mavlink-bidgen if they have modified dialect.
> > > > 
> > > > 
> > > > We can’t really distinguish them since we dont have crc_extra In my opinion it is fine to forward it anyways since whoever really needs to read that packet data will have to parse it
> > > 
> > > 
> > > I disagree, sorry. If you expect different messages there are multiple ways to receive them (like I mentioned above). We cannot expose an insecure API that tries to resolve random payload.
> > 
> > 
> > How would you go about this? I understand your preference for #523 but in some cases, where I use this library, and I really use it a lot, It wouldn't be feasible. I understand if you decide not to merge this or guide it in that direction, I'm just honestly interested in this feature :)
> 
> What is your use case? Can you explain more?

My scenario is basically:

(unknown mavlink definition 1)
third party payload 1
      |
      |serial
     \\/
(known mavlink definition 0) Autopilot ---serial--> CompanionPC (known mavlink definition 0)
     /\\
      |
      |serial
third party payload 2
with unknown mavlink definition 1


So all 3 mavlink generating devices have an unknown mavlink definition where I won't have access because my software in the CompanionPC will just relay information to a specific GCS with that definition

In this scenario i'll only have my autopilot definition but all payloads will be processed by the same application

this happens a lot with third party payloads that produce sensitive information

onur-ozkan:
> > > > > @onur-ozkan
> > > > > > How do you know if it's corrupted packet or valid unknown packet?
> > > > > > Ideally one should simply use #523 or compile it with mavlink-bidgen if they have modified dialect.
> > > > > 
> > > > > 
> > > > > We can’t really distinguish them since we dont have crc_extra In my opinion it is fine to forward it anyways since whoever really needs to read that packet data will have to parse it
> > > > 
> > > > 
> > > > I disagree, sorry. If you expect different messages there are multiple ways to receive them (like I mentioned above). We cannot expose an insecure API that tries to resolve random payload.
> > > 
> > > 
> > > How would you go about this? I understand your preference for #523 but in some cases, where I use this library, and I really use it a lot, It wouldn't be feasible. I understand if you decide not to merge this or guide it in that direction, I'm just honestly interested in this feature :)
> > 
> > 
> > What is your use case? Can you explain more?
> 
> My scenario is basically:
> 
> (unknown mavlink definition 1) third party payload 1 | |serial \/ (known mavlink definition 0) Autopilot ---serial--> CompanionPC (known mavlink definition 0) /\ | |serial third party payload 2 with unknown mavlink definition 1
> 
> So all 3 mavlink generating devices have an unknown mavlink definition where I won't have access because my software in the CompanionPC will just relay information to a specific GCS with that definition
> 
> In this scenario i'll only have my autopilot definition but all payloads will be processed by the same application
> 
> this happens a lot with third party payloads that produce sensitive information

I see. Yeah the existing solutions wouldn't work for your case. But the current diff is not ideal as well.

I have this design in mind:

- Have a unverified message variant/type that contains plain bytes of the incoming message, with a function (somewhere in mavlink-core lib.rs) that reads the message without any verification and returns this variant/type.
- On places where we do verification/validation, if it fails to verify/validate a message, we return the type above inside Err.

Basically, with this approach, you can do whatever you want with the unverified message coming from Err, or you can directly invoke the core reading function (e.g. `fn read_unverified`) from mavlink-core (not from the standard API). So the API remains safe and simple but you can access to the any unverified/unknown messages in multiple ways.

joaofrf:
> > > > > > @onur-ozkan
> > > > > > > How do you know if it's corrupted packet or valid unknown packet?
> > > > > > > Ideally one should simply use #523 or compile it with mavlink-bidgen if they have modified dialect.
> > > > > > 
> > > > > > 
> > > > > > We can’t really distinguish them since we dont have crc_extra In my opinion it is fine to forward it anyways since whoever really needs to read that packet data will have to parse it
> > > > > 
> > > > > 
> > > > > I disagree, sorry. If you expect different messages there are multiple ways to receive them (like I mentioned above). We cannot expose an insecure API that tries to resolve random payload.
> > > > 
> > > > 
> > > > How would you go about this? I understand your preference for #523 but in some cases, where I use this library, and I really use it a lot, It wouldn't be feasible. I understand if you decide not to merge this or guide it in that direction, I'm just honestly interested in this feature :)
> > > 
> > > 
> > > What is your use case? Can you explain more?
> > 
> > 
> > My scenario is basically:
> > (unknown mavlink definition 1) third party payload 1 | |serial / (known mavlink definition 0) Autopilot ---serial--> CompanionPC (known mavlink definition 0) /\ | |serial third party payload 2 with unknown mavlink definition 1
> > So all 3 mavlink generating devices have an unknown mavlink definition where I won't have access because my software in the CompanionPC will just relay information to a specific GCS with that definition
> > In this scenario i'll only have my autopilot definition but all payloads will be processed by the same application
> > this happens a lot with third party payloads that produce sensitive information
> 
> I see. Yeah the existing solutions wouldn't work for your case. But the current diff is not ideal as well.
> 
> I have this design in mind:
> 
> * Have a unverified message variant/type that contains plain bytes of the incoming message, with a function (somewhere in mavlink-core lib.rs) that reads the message without any verification and returns this variant/type.
> * On places where we do verification/validation, if it fails to verify/validate a message, we return the type above inside Err.
> 
> Basically, with this approach, you can do whatever you want with the unverified message coming from Err, or you can directly invoke the core reading function (e.g. `fn read_unverified`) from mavlink-core (not from the standard API). So the API remains safe and simple but you can access to the any unverified/unknown messages in multiple ways.

I'll take a look at that then, and open a PR when I have some spare time. thank you for your comments

PR #531, opened 2026-08-19
Implemented method to retrieve mavlink sha

joaofrf:
New method that allows whoever uses this crate to print out the mavlink commit sha being used

Similar to https://github.com/mavlink/rust-mavlink/pull/530

Something I've had for a while that I believe to be very useful on scenarios where multiple softwares speak mavlink and need to be aligned

### DIFF
--- a/mavlink/src/lib.rs
+++ b/mavlink/src/lib.rs
@@ -84,16 +84,17 @@ pub mod dialects {
     //! - dialect-ualberta
     //! - dialect-uavionix
 
-    /// Git SHA of the MAVLink definitions used to generate the bindings.
+    /// Git SHA of the MAVLink definitions used to generate the bindings, when available.
     ///
-    /// This corresponds to the commit hash in the `mavlink` submodule at build time.
+    /// This is [`None`] for vendored builds and builds performed without Git metadata.
     ///
     /// # Example
     /// ```
-    /// let sha = mavlink::dialects::MAVLINK_DEFINITIONS_SHA;
-    /// println!("Using MAVLink definitions from commit: {sha}");
+    /// if let Some(sha) = mavlink::dialects::MAVLINK_DEFINITIONS_SHA {
+    ///     println!("Using MAVLink definitions from commit: {sha}");
+    /// }
     /// ```
-    pub const MAVLINK_DEFINITIONS_SHA: &str = env!("MAVLINK_SHA");
+    pub const MAVLINK_DEFINITIONS_SHA: Option<&str> = option_env!("MAVLINK_SHA");
 
     include!(concat!(env!("OUT_DIR"), "/mod.rs"));
 }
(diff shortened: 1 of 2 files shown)


joaofrf:
Fixed formatting

pv42:
I like the idea of this but I believe it would break on vendored builds (e.g. https://crates.io/api/v1/crates/mavlink/0.17.1/download) and on machienes without git installed.

onur-ozkan:
> I like the idea of this but I believe it would break on vendored builds (e.g. https://crates.io/api/v1/crates/mavlink/0.17.1/download) and on machienes without git installed.

Yes

mavlink/src/lib.rs
```diff
@@ -87,3 +87,16 @@ pub mod dialects {
...
+/// Returns the git SHA of the MAVLink definitions used to generate the bindings.
+///
+/// This corresponds to the commit hash in the `mavlink` submodule at build time.
+///
+/// # Example
+/// ```
+/// let sha = mavlink::mavlink_definitions_sha();
+/// println!("Using MAVLink definitions from commit: {sha}");
+/// ```
+pub const fn mavlink_definitions_sha() -> &'static str {
```
onur-ozkan: Could you please move this under the dialects module? Also, a const variable seems like a better fit for this.

We should also find a way that works with vendor builds.

joaofrf: i'll move it under the dialects module
I've no idea how to tackle vendor builds, any idea? (that doesn't use hardcoded or a file containing the sha)

joaofrf: We can make the SHA unavailable when Git is missing

joaofrf: For now i've changed it to:
Option and it is now calculated inside is_submodule condition
I believe this is enough to avoid breaking vendor builds

mavlink/src/lib.rs
```diff
@@ -83,6 +83,19 @@ pub mod dialects {
     //! - dialect-test
     //! - dialect-ualberta
     //! - dialect-uavionix
+
+    /// Git SHA of the MAVLink definitions used to generate the bindings, when available.
+    ///
+    /// This is [`None`] for vendored builds and builds performed without Git metadata.
```
onur-ozkan:
```suggestion
    /// This is [`None`] when `git` fails to resolve the commit SHA.
```

mavlink/src/lib.rs
```diff
@@ -83,6 +83,19 @@ pub mod dialects {
...
+    ///
+    /// # Example
```
onur-ozkan: I think we can remove the example section easily.

onur-ozkan (commented):
LGTM other than 2 nits:

mavlink/src/lib.rs
joaofrf: agreed, i've pushed the changes

PR #535, opened 2026-09-03
Replace peek parsing with an SIMD powered incremental reader

onur-ozkan:
The existing reader requests the marker, header, and body separately through `PeekReader`. This results in repeated small reads and byte-by-byte scanning whenever the stream is not synchronized.

Changes in this PR:

- Stores partial frames and read-ahead data in one reusable buffer so transports no longer need an additional `BufReader`.
- Reads across frame boundaries to allow one source read to supply multiple MAVLink frames which reduces calls to the underlying transport syscalls.
- Parses messages directly from buffered frame slices.
- Retains incomplete input across I/O errors and asynchronous cancellation.
- Shares framing and CRC-validation logic across `std`, Tokio and embedded readers.
- Removes `PeekReader`, `AsyncPeekReader` and the duplicated parsing paths.

Marker lookup first checks the byte at the current cursor. Synchronized streams therefore stay on the scalar fast path. After a miss, the remaining buffered input is searched using SIMD where supported (SIMD speeds up significantly as noise increases). Invalid candidates are rejected one byte past the candidate marker, so valid frames embedded in corrupted input are preserved.

#### Benchmark

The benchmark replays the static PX4 `mavlink/tests/log.tlog` five times in memory. See the [benchmark code](https://github.com/Orkavian/rust-mavlink/commit/93c0846e6ae44aa325b24836678dc3d7d89e3eef).

```sh
$ cargo bench --manifest-path benchmarks/file_reader/Cargo.toml
   Compiling file-reader-benchmark v0.0.0 (/rust-mavlink/benchmarks/file_reader)
    Finished `bench` profile [optimized] target(s) in 0.27s
     Running benches/file_reader.rs (benchmarks/file_reader/target/release/deps/file_reader-892f94bcbb74346f)
stream    api       reader      messages        time     throughput  vs master
--------  --------  --------  ----------  ----------  -------------  ---------
clean     blocking  master          7130    1.294 ms      236.1 MiB/s          -
clean     blocking  current         7130    0.474 ms      645.2 MiB/s      2.73x
clean     tokio     master          7130    1.926 ms      158.7 MiB/s          -
clean     tokio     current         7130    0.605 ms      504.8 MiB/s      3.18x
noisy     blocking  master          7130    1.526 ms      226.2 MiB/s          -
noisy     blocking  current         7130    0.499 ms      691.9 MiB/s      3.06x
noisy     tokio     master          7130    2.425 ms      142.4 MiB/s          -
noisy     tokio     current         7130    0.629 ms      548.6 MiB/s      3.85x
```

The clean stream replays the capture unchanged. The noisy stream preserves every frame but inserts deterministic 5–30-byte noise bursts between runs of 1–5 frames. Noise accounts for roughly 10% of the stream. Both readers successfully decode all 7,130 messages.


onur-ozkan:
~I will try to write proper benchmark code and share it. Until then the PR will remain draft.~ DONE.

But I would like to get some early reviews as possible!

onur-ozkan:
Could one of you take a look at this when you have some time? @joaoantoniocardoso @patrickelectric

onur-ozkan:
Please let me know if you don't have any major concerns so I can land this. @patrickelectric

patrickelectric:
Can I take a look by tomorrow ?

onur-ozkan:
> Can I take a look by tomorrow ?

Of course, thanks!

patrickelectric:
Reviewing!

image

mavlink-core/src/reader.rs
```diff
@@ -0,0 +1,443 @@
...
+            message.0[..frame.bytes().len()].copy_from_slice(frame.bytes());
+            MAVLinkMessageRaw::V1(message)
+        }
+        MavlinkVersion::V2 => {
+            let mut message = MAVLinkV2MessageRaw::new();
+            message.0[..frame.bytes().len()].copy_from_slice(frame.bytes());
+            MAVLinkMessageRaw::V2(message)
+        }
+    }
+}
```
patrickelectric: can't we just create a owned_v2 and owned_v1 to remove the unreachable calls ?

```suggestion
#[inline]
fn owned_v1_raw_message(frame: FrameRef<'_>) -> MAVLinkV1MessageRaw {
    let mut message = MAVLinkV1MessageRaw::new();
    message.0[..frame.bytes().len()].copy_from_slice(frame.bytes());
    message
}

#[inline]
fn owned_v2_raw_message(frame: FrameRef<'_>) -> MAVLinkV2MessageRaw {
    let mut message = MAVLinkV2MessageRaw::new();
    message.0[..frame.bytes().len()].copy_from_slice(frame.bytes());
    message
}

#[inline]
fn owned_raw_message(frame: FrameRef<'_>) -> MAVLinkMessageRaw {
    match frame.version() {
        MavlinkVersion::V1 => MAVLinkMessageRaw::V1(owned_v1_raw_message(frame)),
        MavlinkVersion::V2 => MAVLinkMessageRaw::V2(owned_v2_raw_message(frame)),
    }
}
```

and if that we could do:

```rs
MavlinkVersion::V2 => {
    signing_data.verify_signature(&owned_v2_raw_message(frame))
}
```

mavlink/tests/signing.rs
```diff
@@ -50,8 +50,13 @@ mod signing {
...
         let signing_cfg = SigningConfig::new(SECRET_KEY, 0, true, false);
         let signing_data = SigningData::from_config(signing_cfg);
-        let mut r = PeekReader::new(HEARTBEAT_SIGNED);
-        let msg = read_v2_raw_message::<mavlink::dialects::common::MavMessage, _>(&mut r).unwrap();
+        let mut r = MavlinkReader::new(HEARTBEAT_SIGNED);
+        let MAVLinkMessageRaw::V2(msg) = r
+            .read_raw_message::<mavlink::dialects::common::MavMessage>(MavlinkVersion::V2)
+            .unwrap()
+        else {
+            unreachable!()
```
patrickelectric, line 58: can't we do the same thing with v1 and v2 to avoid this unreachable ?

patrickelectric (changes requested):
We are missing a test in CI for:
`cargo test --verbose --features tokio,mav2-message-signing`

mavlink-core/src/reader.rs
onur-ozkan: >can't we just create a owned_v2 and owned_v1 to remove the unreachable calls ?

Where is unreachable call in this  code? I think current code looks better than creating v1 and v2 specific functions, you may end up sending invalid frames to those calls as well.

mavlink/tests/signing.rs
onur-ozkan: We can avoid unreachable by directly calling `owned_raw_message` instead of matching on `frame.version()`, but that would do unnecessary work for v1 frames.

Or, we can embed what `owned_raw_message` is doing for v2.

Tbh, I don't see any harm on unreachable because just couple lines above the code enters v2, so if we ever end up calling `unreachable` then we have serious bug.

onur-ozkan:
> We are missing a test in CI for: `cargo test --verbose --features tokio,mav2-message-signing`

Added now.

mavlink-core/src/reader.rs
patrickelectric: https://github.com/mavlink/rust-mavlink/pull/535/changes#diff-d420a98497be6c6ee322d0c86ba94cbf5fb94ae45578398d4edb34d1d138709fR311

mavlink/tests/signing.rs
patrickelectric: > end up calling unreachable then we have serious bug

This should not happen in a library that's kind of critical. If there is any possibility to fail, we should return something for the user to take care of it.

My suggestion is to follow that I described in the previous comment about `owned_raw_message`, it's just moving some inner logic to shared function for the code. It also removes the unreachable completely. If the suggested code **can** fail and unreachable is a possibility, we should deal with that over calling unreachable.

onur-ozkan: I do agree for some points but the proposed code looks iffy to me. Can you check [this alternative](https://github.com/mavlink/rust-mavlink/pull/535/commits/3bbb4d5c6a423e33968515a45388f5340baec630)?

patrickelectric: looks good

onur-ozkan: The unreachable call from the library is removed, this one is used in tests, which should be fine, right?

onur-ozkan:
Do we have any blockers?

PR #542, opened 2026-09-18
Add mavlink sha to generated dialect file docs

pv42:
This modifies the top comment of dialect files to to include the mavlink sha that they were generated from (if available). 
e.g.: 

MAVLink common dialect.

This file was automatically generated from [mavlink @ 44a1229ece43cfc3ce1640451ad5aaba52a28459](https://github.com/mavlink/mavlink/tree/44a1229ece43cfc3ce1640451ad5aaba52a28459), do not edit."


mavlink/build/main.rs
```diff
@@ -119,7 +126,7 @@ fn main() -> ExitCode {
     };
 
     let out_dir = env::var("OUT_DIR").unwrap();
-    let result = match mavlink_bindgen::generate(xml_definitions, out_dir) {
+    let result = match mavlink_bindgen::generate(xml_definitions, out_dir, mavlink_sha.as_deref()) {
```
onur-ozkan, line 129: I think we should add proper versioning to the bindgen. For example, one can calculate their own hash (from their dialect file?) and use it, but then the comment will be invalid (as it points to the mavlink repository).

I don't have a design for this "proper versioning" mechanism but it shouldn't be too complicated. We just need to add a way to add version information to the dialects that isn't necessarily related to the submodules or git. Maybe we can add specific information on the hash e.g. `Version::new("File hash", $hash)`, `Version::new("mavlink sha", $hash)`, etc.

onur-ozkan: Also, I think it should be exposed as a variable as well (e.g., a global const) so it can be accessed programmatically rather than only being mentioned in a doc comment in the file.

pv42: I added a const to the generated files so it can be accessed programmatically:
```rs
pub const MAVLINK_SHA: &str = "44a1229ece43cfc3ce1640451ad5aaba52a28459";
```

onur-ozkan: Nice, thanks! We still need a generic versioning mechanism. Like I said I don't have clear design on mind but it should be generic to handle every use case with least amount of complexity. Perhaps something like this type could help

```
struct Version {
    dialect_file_metadata 
    mavlink_sha (optional)
}
```

pv42: > Nice, thanks! We still need a generic versioning mechanism. Like I said I don't have clear design on mind but it should be generic to handle every use case with least amount of complexity. Perhaps something like this type could help
> 
> ```
> struct Version {
>     dialect_file_metadata 
>     mavlink_sha (optional)
> }
> ```

Alright converting this to draft until I have done that.

